PR #5, opened 2022-06-02
README: Add more details

oceanc80:
Expand README and add templated sections to be filled in later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,119 @@
 # deppy
+
+[![License](http://img.shields.io/:license-apache-blue.svg)](http://www.apache.org/licenses/LICENSE-2.0.html)
+[![Go Report Card](https://goreportcard.com/badge/github.com/operator-framework/deppy)](https://goreportcard.com/report/github.com/operator-framework/deppy)
+[![Slack Channel](https://img.shields.io/badge/chat-4A154B?logo=slack&logoColor=white "Slack Channel")](https://kubernetes.slack.com/archives/C038B7MF75M)
+
 Deppy: The dependency resolver for Kubernetes
+
+## Introduction
+
+Deppy is a Kubernetes API that runs on- or off-cluster for resolving constraints over catalogs of
+[RukPak](https://github.com/operator-framework/rukpak) bundles.
+Deppy is part of the next iteration of OLM and was first introduced
+[here](https://hackmd.io/upiNuoeJTwqNKQJCVMZADw).
+The initial goal of the project is to remove the dependency manager from the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,2 +1,38 @@
...
+
+## Introduction
+
+Deppy is a Kubernetes API for resolving constraints over
+[RukPak](https://github.com/operator-framework/rukpak) bundle catalog(s).
+Deppy is part of the next iteration of OLM and was first introduced
+[here](https://hackmd.io/upiNuoeJTwqNKQJCVMZADw).
+The initial goal of the project is to remove the dependency manager from the
+[Operator Lifecycle Manager](https://github.com/operator-framework/operator-lifecycle-manager)
+(OLM) and make it its own generic component.
```
timflannagan, line 17: A couple of comments here:

- There's no concept of rukpak bundle catalog(s). Any package-manager-related concepts are applicable to deppy (e.g. sourcing catalogs, indices, registeries, etc.).
- The initial project goal seems a fine thing to have as a first pass

Maybe we can use the e2e strawman's deppy introduction language:

> Deppy runs on- or off-cluster to provide dependency resolution for catalogs of RukPak bundles

Thoughts?

exdx (approved):
LGTM to me for an initial README, but it would be nice to include some high-level descriptions and examples that could help guide the user looking at deppy for the first time (since it's pretty abstract and less intuitive than something like rukpak):
* What is a constraint in this context (with an example)?
* What is resolution in this context?

README.md
oceanc80: Sounds good. I just copied the intro from [this doc](https://hackmd.io/sxT8wkBvTIy1n9C27Eo7SQ?view). Should we change that there as well?

timflannagan: Yeah I think so, but my feedback doesn't have to be blocking. We can merge this as-is and handle any follow-ups over time.

README.md
```diff
@@ -1,2 +1,118 @@
...
+
+### Concepts
+
+Within the context of a package manager, there are a set of expectations from the
+user that the package manager should never:
+
+- install a package whose dependencies cannot be fulfilled or that conflict with
+another package's dependencies
+- update a package in a way that breaks another that depends on it
```
exdx, line 26: Might also be helpful to add a case around constraints here:
- Install a package whose constraints cannot be met by the current set of installable packages

exdx (commented):
LGTM

tylerslaton (approved):

PR #6, opened 2022-06-03
api,controllers: Add barebones Input API

timflannagan:
Update the api/v1alpha1/input_types.go and update the InputSpec to match the [e2e strawman's proposed API surface](https://hackmd.io/hppR60SiRGKPcqKQuByvVg#Input-API).

Signed-off-by: timflannagan <timflannagan@gmail.com>


api/v1alpha1/input_types.go
```diff
@@ -17,25 +17,27 @@ limitations under the License.
...
+	Constraints    []Constraint `json:"constraints,omitempty"`
+}
 
-	// Foo is an example field of Input. Edit input_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+type Constraint struct {
+	Type string `json:"type"`
+	//+kubebuilder:validation:Schemaless
+	//+kubebuilder:validation:XPreserveUnknownFields
+	Value json.RawMessage `json:"value"`
```
tylerslaton, line 35: What would be an example of the value here?

timflannagan: I think there's a bunch of different values here. You could have a type/value constraint where the value is just a string, or the value is more generic and contains a custom structure:

```
constraints:
- type: olm.package
  value: prometheus
- type: olm.channel
  value: "4.12"
- type: olm.catalog
  value:
    name: redhat-operators
    namespace: openshift-marketplace
...
```

tylerslaton: Got it. That implies we have some underlying logic in the controller that understands how to read each individual type, right?

timflannagan: That's certainly an open question. I think we could define a set of deppy constraints out-of-the-box but still leave the API pluggable in some capacity. Maybe we just restrict constraints to be a string value, and build on top of that over time. Thoughts @joelanford?

joelanford: I have a POC that makes all of the constraint types pluggable via a Kubernetes API with inlined javascript. Then we'd just include the "built-in" constraint types as objects in the deppy install.

api/v1alpha1/input_types.go
```diff
@@ -17,25 +17,27 @@ limitations under the License.
...
 }
 
 // InputStatus defines the observed state of Input
 type InputStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
tylerslaton, line 40: Would it make sense to have the `solution` ([kuberdep](https://github.com/joelanford/kuberdep/blob/cbadc7a88b502b167afb304d61f2bba03e5b0c03/api/v1/resolution_types.go#L33)) or `instances` ([strawman](https://hackmd.io/hppR60SiRGKPcqKQuByvVg?view#Input-API)) property inside of the status here to define the result of a resolution? I know this is a barebone approach so this might make sense to address in a follow-up.

```suggestion
	Conditions []metav1.Condition `json:"conditions,omitempty"`
	Solution []string `json:"solutions,omitempty"`
```

tylerslaton (approved):

PR #7, opened 2022-06-03
ci: Add golangci-lint as a tool dependency

timflannagan:
Update the Makefile and introduce the golangci-lint target to ensure
that local/CI environments use the same version of the golangci-lint
tool.

Update the sanity workflow and ensure CI is running the `lint` Makefile
target, which runs that golangci-lint target on the repository path.

Because we're replacing the golangci-lint github action with running a
Makefile target, we need to cache the go build/module paths so we don't
increase the overall CI sanity workflow time as that
action was previously handling that for us under-the-hood.

---

Update the root .golangci-lint configuration and enforce the
rukpakv1alpha1 and deppyv1alpha1 package import naming.

---

Fix any linting violations after updating that importas golangc-lint
configuration to enforce deppyv1alpha1 as the package import name.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -30,6 +30,10 @@ linters-settings:
       alias: metav1
     - pkg: k8s.io/apimachinery/pkg/api/errors
       alias: apierrors
+    - pkg: github.com/operator-framework/rukpak/api/v1alpha1
+      alias: rukpakv1alpha1
+    - pkg: github.com/operator-framework/deppy/api/v1alpha1
+      alias: deppyv1alpha1
   goimports:
     local-prefixes: github.com/operator-framework/deppy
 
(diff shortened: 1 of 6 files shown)


.golangci.yaml
```diff
@@ -30,6 +30,10 @@ linters-settings:
       alias: metav1
     - pkg: k8s.io/apimachinery/pkg/api/errors
       alias: apierrors
+    - pkg: github.com/operator-framework/rukpak/api/v1alpha1
```
tylerslaton, line 33: nit: Copy and paste artifact?

tylerslaton (approved):
Non-blocking nit, otherwise 🚀

.golangci.yaml
timflannagan: Nope: I think we'll interface with rukpak resources in the input controller at some point.

PR #18, opened 2022-06-13
api: Add Property to Input; add k/v Values

### LINKED ISSUE #10: Update the Input API and separate property and constraint definitions
Goal: update the Input API and introduce .Spec fields for property and constraint definitions.

AC:

- Add a spec.Properties type definition where an individual property is a string key/value pair
- Update the spec.Constraints type definition to also have a string key/value pair type

exdx:
Adds a new Property type to Input.spec. 
Updates the Value type to be a key/value pair, modeled as a map[string]string.

The API now becomes something like:
```
apiVersion: core.deppy.io/v1alpha1
kind: Input
metadata:
  name: input-sample
spec:
  inputClassName: "sub"
  constraints:
    - type: deppy.package.required
      value:
        packageName: baz
        versionRange: '>=1.0.0'
  properties:
    - type: deppy.gvk.provided
      value:
        group: operators.coreos.com
        version: v1
        kind: foo

```
  

Closes #10


oceanc80 (approved):

tylerslaton (approved):
🚀

exdx:
Merging as-is -- we can always make a change to the API in the near term as we continue iterating.

PR #19, opened 2022-06-21
DO NOT MERGE: add a test demonstrating non-working complex constraint

joelanford:
I'm experimenting with adding support in the solver package for a compound constraint type. For example:

```go
solver.AllOf(solver.Dependency("x3", "x2"), solver.Dependency("x3", "x1"))
```

I might be thinking about this incorrectly, but it seems like that should translate to a logic circuit that looks like:
```
!subject || ( (x3 || x2) && (x3 || x1) )
```

This would mean that if `subject` is mandatory, then the only satisfiable solution would be `[subject, x3]`. However, this PR adds a test that fails on this example.


internal/solver/solve_test.go
```diff
@@ -363,3 +395,93 @@ func TestDuplicateIdentifier(t *testing.T) {
...
+	x2 := lm.LitOf("x2")
+	x3 := lm.LitOf("x3")
+
+	dep := c.Or(subLit.Not(), c.Ands(c.Or(x3, x2), c.Or(x3, x1), c.F))
+	fmt.Printf("!%s || ((x3 || x2) && (x3 || x1) && FALSE)\n", subject)
+	return dep
+}
+
+func (constraint hardcodedUNSAT) order() []Identifier {
+	return nil
```
benluddy, line 444:
```suggestion
	return []Identifier{"x3", "x2", "x1"}
```

PR #26, opened 2022-06-30
Update tooling

### LINKED ISSUE #21: Update tooling to mirror rukpak's setup
The tooling for deppy is similar but different than rukpak (where there is a hack/tools with a separate go.mod). It might be worth updating the tooling now, early on in the project, to mirror rukpak, so there is a consistent way to use tools between the two projects.

tylerslaton:
Doing three things here:

1. Adding a tools directory in the hack folder ala [Rukpak](https://github.com/operator-framework/rukpak/tree/main/hack/tools).
2. Updating the Makefile target to more closely align with how we have it defined [in Rukpak](https://github.com/operator-framework/rukpak/tree/main/Makefile).
3. Updating the Github actions to utilize the updating Makefile targets.

All three of these changes are in their own commits for ease of review. I will squash them together we're ready to merge.

Closes #21


timflannagan (commented):
I may need some convincing here. While I can appreciate rukpak's tooling layout, it feels like we should either use a go workspace _or_ just continue to `go install ...` the tooling dependencies and tracking the individual tool version through Makefile versions. Either approach should still result in the same local/CI environment experience.

tylerslaton:
@timflannagan I'm fine with either the Rukpak tooling methodology or the default generated one that Deppy is using. In either case I feel that we should make them consistent across each just for ease of use. Thoughts?

Makefile
```diff
@@ -57,118 +68,129 @@ fmt: ## Run go fmt against code.
...
+controller-gen: $(CONTROLLER_GEN) ## Build a local copy of controller-gen
+ginkgo: $(GINKGO) ## Build a local copy of ginkgo
+golangci-lint: $(GOLANGCI_LINT) ## Build a local copy of golangci-lint
+setup-envtest: $(SETUP_ENVTEST) ## Build a local copy of envtest
+goreleaser: $(GORELEASER) ## Builds a local copy of goreleaser
+kind: $(KIND) ## Builds a local copy of kind
+kustomize: $(KUSTOMIZE) ## Builds a local copy of kustomize
+
+$(KUSTOMIZE): $(TOOLS_BIN_DIR) ## Build kustomize to the tools binary folder
+	curl -s $(KUSTOMIZE_INSTALL_SCRIPT) | bash -s -- $(subst v,,$(KUSTOMIZE_VERSION)) $(TOOLS_BIN_DIR)
```
akihikokuroda, line 184: This line fails when `make run` is executed the second time.  The rukpak is using `kubectl -k` instead of the standalone kustomize.   Would you make changes not using the standalone kustomize or just add "-" in front of theh`curl` so that the error is ignored?

PR #35, opened 2022-10-13
Add Entity structure definition

perdasilva:
This PR adds the `Entity` definition as a structure with an `Id` and `Properties` (string->string).
It also adds ginkgo and gomega to the project and uses them for the unit tests.


timflannagan (approved):
lgtm once the linting job passes

internal/source/entity.go
```diff
@@ -0,0 +1,35 @@
...
+func NewEntity(id EntityID, properties map[string]string) *Entity {
+	return &Entity{
+		id:         id,
+		properties: properties,
+	}
+}
+
+func (e *Entity) ID() EntityID {
+	return e.id
+}
```
awgreene, line 27: Was this method created to address the concern that the ID could be modified after creation? Should we explicitly call out that the field is immutable as a comment on the structure?

awgreene (approved):
Left a nit, but this looks good otherwise.
/lgtm

internal/source/entity.go
timflannagan: The underlying `id` field in the Entity structure is unexported, so it should be immutable after creation of this type given there's no setter-esq method available. This method is just exposing a way to get the value of the underlying id field.

internal/source/entity_test.go
```diff
@@ -0,0 +1,25 @@
+package source_test
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/operator-framework/deppy/internal/source"
```
timflannagan, line 6: The linter is complaining about this grouping. It wants the internal package to be declared separate from the third-party packages (e.g. a newline between line 5 and line 6).

internal/source/entity.go
perdasilva: Right, I was just protecting the data of the entity from mutation

PR #37, opened 2022-10-14
Add Entity universe

perdasilva:
The entity universe represents the complete set of entities over which deppy will operate. The entity universe surfaces methods for searching and sorting entities. The entity universe gives constraint factories the access to the entities in a queryable form, for instance:

```go
func packageDependency(subject solver.Identifier, packageName string, versionRange string) []solver.Constraint {
  ids := entityUniverse.Search(universe.And(withPackageName(packageName), withinVersion(versionRange))).Sort(byChannelAndVersion).CollectIds()
  return []solver.Constraint{solver.Dependency(subject, ids...)}
}
```

where `withPackageName`, `withinVersion`, `byChannelAndVersion` are `Predicate` and `SortFunction` implementations

### DIFF
--- a/internal/universe/entity_universe.go
+++ b/internal/universe/entity_universe.go
@@ -1,0 +1,46 @@
+package universe
+
+import (
+	"github.com/operator-framework/deppy/internal/source"
+)
+
+// TODO: https://github.com/operator-framework/deppy/issues/40
+
+// EntityUniverse represents a queryable set of entities
+// it is used as a basis for creating solver constraints
+type EntityUniverse struct {
+	universe []source.Entity
+	idIndex  map[source.EntityID]source.Entity
+}
+
+func NewEntityUniverse(entities []source.Entity) *EntityUniverse {
+	m := make(map[source.EntityID]source.Entity)
+	for _, entity := range entities {
+		m[entity.ID()] = entity
+	}
+	return &EntityUniverse{
+		universe: entities,
+		idIndex:  m,
+	}
+}
+
+func (u *EntityUniverse) Get(id source.EntityID) *source.Entity {
+	if entity, ok := u.idIndex[id]; ok {
+		return &entity
+	}
+	return nil
+}
+
+func (u *EntityUniverse) Search(predicate Predicate) SearchResult {
+	out := make([]source.Entity, 0)
+	for _, entity := range u.universe {
+		if predicate(&entity) {
+			out = append(out, entity)
+		}
+	}
+	return out
+}
+
+func (u *EntityUniverse) AllEntities() SearchResult {
+	return u.universe
+}
--- a/internal/universe/entity_universe_test.go
+++ b/internal/universe/entity_universe_test.go
@@ -1,0 +1,51 @@
+package universe_test
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/operator-framework/deppy/internal/source"
+	"github.com/operator-framework/deppy/internal/universe"
+)
+
+var _ = Describe("EntityUniverse", func() {
+	var entityUniverse = universe.NewEntityUniverse([]source.Entity{
+		*source.NewEntity("1", map[string]string{"property": "1"}),
+		*source.NewEntity("2", map[string]string{"property": "2"}),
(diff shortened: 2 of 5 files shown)


internal/universe/entity_universe.go
```diff
@@ -0,0 +1,42 @@
+package universe
+
+import (
+	"github.com/operator-framework/deppy/internal/source"
+)
+
+type EntityUniverse struct {
+	universe []source.Entity
```
timflannagan, line 8: This can be out-of-scope for now: do we need to configure a max size of the universe, or add some sort of eviction mechanisms?

internal/universe/entity_universe.go
```diff
@@ -0,0 +1,42 @@
...
+	return nil
+}
+
+func (u *EntityUniverse) Search(predicate Predicate) SearchResult {
+	out := make([]source.Entity, 0)
+	for _, entity := range u.universe {
+		if predicate(&entity) {
+			out = append(out, entity)
+		}
+	}
```
timflannagan, line 36: I wonder whether the universe could be sharded/namespaced to make client-side searching a bit easier vs. relying on predicate filtering but that's something we can revisit over time.

internal/universe/entity_universe.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/operator-framework/deppy/internal/source"
+)
+
+type EntityUniverse struct {
+	universe []source.Entity
+	idIndex  map[source.EntityID]source.Entity
+}
+
+func NewEntityUniverse(entities []source.Entity) *EntityUniverse {
```
timflannagan, line 12: What happens if the universe needs to be updated? It doesn't look like there's any way of modify the current set of the underlying universe without re-creating a new EntityUniverse. Is that the right pattern, or do we need a way to dynamically resize/update the universe?

internal/universe/entity_universe.go
```diff
@@ -0,0 +1,42 @@
+package universe
+
+import (
+	"github.com/operator-framework/deppy/internal/source"
+)
+
+type EntityUniverse struct {
```
timflannagan, line 7: not blocking: add a godoc for the EntityUniverse type?

timflannagan:
Could we add an example of how we'd expect this library to be used during any constraint building operations if you had already have something in mind?

internal/universe/entity_universe_test.go
```diff
@@ -0,0 +1,51 @@
...
+	})
+
+	It("should be searchable", func() {
+		entities := entityUniverse.Search(func(entity *source.Entity) bool {
+			value, err := entity.GetProperty("property")
+			if err != nil {
+				return false
+			}
+			return value <= "2"
+		})
```
timflannagan, line 35: Nice :+1:

akihikokuroda (approved):
LGTM

internal/universe/entity_universe.go
perdasilva: yeah - at the moment this is really barebones, just to get some traction. We need to think a little bit about the query api and how that will play with the deppy sources. I think in a production setting, the collection of deppy sources will be a database and we'll have a client that to query everything. Over time we can evolve the implementation to not, e.g. requery everything, but rather keep the queries cached and watch for changes in the db and update the result set.

timflannagan: Yeah, that's fine with me. Do we need to track anything in the short term then to make sure we're capturing any context while we have a barebones implementation layer?

internal/universe/entity_universe.go
perdasilva: Yeah, we'll definitely want to make the set of sources queryable as well.

timflannagan: This is fine for a first pass implementation so I'm cool with merging this, and iterating on top of it.

internal/universe/entity_universe.go
perdasilva: https://github.com/operator-framework/deppy/issues/40 - I've created an issue and linked it in code as well

internal/universe/entity_universe.go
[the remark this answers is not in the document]
perdasilva: added to the issue: https://github.com/operator-framework/deppy/issues/40

PR #38, opened 2022-10-14
[WIP] docs: add document describing the ActiveResolution API

tylerslaton:
> **Note**: Moving the conversation for this document to be in-repo and the deliverable to be committed in-tree.

### Relevant documents
- https://hackmd.io/9LfeLfLkTnazoO0Mpjm5Yw

### Relevant slack conversations
- https://kubernetes.slack.com/archives/C045L6WK27N/p1665600917035019
- https://kubernetes.slack.com/archives/C045L6WK27N/p1665598035700839
- https://kubernetes.slack.com/archives/C045L6WK27N/p1665598022439249

### DIFF
--- a/docs/concepts/active-resolution.md
+++ b/docs/concepts/active-resolution.md
@@ -1,0 +1,126 @@
+# Active Resolution
+
+The ActiveResolution API and its controller will manage the state of ActiveResolution resources. It executes the main algorithm of producing solver constraints from constraint definitions and entities, triggers resolution, converts results to something consumable by users and writes results out to the particular ActiveResolution resource.
+
+## API definition
+
+Though not the final definition for this resource, the following provides an example of what it might look like:
+
+```yaml
+apiVersion: deppy.io/v1alpha1
+kind: ActiveResolution
+metadata:
+  name: example-resolution
+spec:
+  sources: # optional definition of specific sources
+  - "example-operators"
+  constraints: # define constraints necassary for selections
+  - required("first-example-operator", ">= 1.0.0", "stable")
+  # or
+  - required:
+      name: "second-example-operator"
+      versionRange: ">= 1.0.0 < 1.2.0"
+      # optionally
+      channel: "stable"
+status:
+  conditions: # status conditions
+  ...
+  selections: # items selected in the form of source:name:version:channel
+  - example-operators:first-example-operator:1.2.0:stable
+  - example-operators:second-example-operator:1.1.0:stable
+  - ...
+```
+
+## Functionality
+
+For consumers of `Deppy`, they will be expected to provide the `sources` to reach out to as well as the `constraints` necessary for the desired resolution. Using this information, `Deppy` will then reach out to the appropriate cluster resources to generate a resolution. **`Deppy` will not be responsible for installing this content onto the cluster**, instead it is up to the consumer to determine how to act on a `ActiveResolution`'s status.
+
+### With the Operator API
+
+Its important to illustrate how the `Operator API` use case can utilize `Deppy` to understand the ultimate contract between the services for Phase 0 of `Deppy`. 
+
+> **Note**: Keep in mind that the `Operator API` has not been completely designed. For the sake of discussing how the `ActiveResolution` component works within the larger OLM V1 system, lets assume that this `Operator API` has been designed. It takes has a single crd called `Operator` which can take in a package to install.
+
+Say that we have defined an example `Operator` like so:
+
+```yaml
+apiVersion: platform.openshift.io/v1alpha1
+kind: Operator
+metadata:
+  name: foo-operator
+spec:
+  package:
+    name: foo-package
+```
+
+Upon applying this resource, there are two API's that will now be at play - Deppy and RukPak. To start the process, the `Operator` controller will create an `ActiveResolution` for `Deppy` to reconcile. 
+
+```yaml
+apiVersion: deppy.io/v1alpha1
+kind: ActiveResolution
+metadata:
+  name: foo-resolution
+spec:
+  constraints:
+  - required:
+      name: "foo-package"
+      channel: "platform"
+...
+```
+
+For the sake of example, lets say that `foo-package` is reliant on another package called `bar-package`. In this case, `Deppy` will understand and convey this dependency to the status.
+
+```yaml
+...
+status:
+  conditions: # status conditions
+  ...
+  selections: # items selected in the form of source:name:version:channel
+  - redhat-operators:foo-package:1.2.0:platform
+  - redhat-operators:bar-package:1.1.0:platform
+```
+
+Looking at the status, we can see that there is a constructed query that tells us where our resolved entitiy is located at. The `Operator` controller could read this and perform another query to get the images; However, it may be pertitnent for a new component to be introduced into Deppy - the `ResolutionActivator` (name TBD). 
+
+### Resolution Activator
+
+The responsibility for the `ResolutionActivator` is to be able to read `ActiveResolution`s and parse out the installable units of work. This could follow an adapter pattern, like Deppy itself does,  where the adapters are able to retrieve installable content. In the use case of `Operator`s this would result in a collections of images being made available. To mark a `ActiveResolution` with the appropriate adapter as well as if it should be activated, we can use annotations.
+
+```yaml
+apiVersion: deppy.io/v1alpha1
+kind: ActiveResolution
+metadata:
+  annotations:
+    deppy.io/v1alpha1/activator-class: "Operator"
+  name: foo-resolution
+...
+```
+
+This would result in the status being updated to include a new section - entities.
+
+```yaml
+...
+status:
+  conditions: # status conditions
+  ...
+  selections: # items selected in the form of source:name:version:channel
+  - redhat-operators:foo-package:1.2.0:platform
+  - redhat-operators:bar-package:1.1.0:platform
+  entities: # entities that were made installable via the activator-class annotation
+  - package: foo-package
+    image: quay.io/redhat-operators/foo-package:v1.2.0
+  - package: bar-package
+    image: quay.io/redhat-operators/bar-package:v1.1.0
+```
+
+From here, the `Operator` controller can read the resolved `entities` and install them accordingly onto the cluster via `RukPak`. However, it is pertinent to allow `Rukpak` to manage these installables as a group of installed content instead of individuals. This is where the `ResolveSet` bundle format comes into play. 
+
+A `ResolveSet` bundle is, at its core, just a `plain+v0` bundle of `BundleDeployments`. What this means is that when the `ResolveSet` bundle gets installed the `plain` provisioner will work to reconcile that content as though it were arbitrary manifests, essentially having a wrapper `BundleDeployment` for `BundleDeployments`. Generation of the `ResolveSet` bundle is trivial and can be done on the fly by the `Operator` controller. If we were to look at a `ResolveSet` bundle's manifests for the above example, it would look something like this:
+
+```console
+manifests/
+|- foo-package-bd.yaml
+|- bar-package-bd.yaml
+```
+
+From this, the `Operator` controller is able to maintain a single `BundleDeployment` for all of the resolved content.


docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
...
+  entities: # entities that were made installable via the activator-class annotation
+  - package: foo-package
+    image: quay.io/redhat-operators/foo-package:v1.2.0
+  - package: bar-package
+    image: quay.io/redhat-operators/bar-package:v1.1.0
+```
+
+From here, the `Operator` controller can read the resolved `entities` and install them accordingly onto the cluster via `RukPak`. However, it is pertinent to allow `Rukpak` to manage these installables as a group of installed content instead of individuals. This is where the `ResolveSet` bundle format comes into play. 
+
+A `ResolveSet` bundle is, at its core, just a `plain+v0` bundle of `BundleDeployments`. What this means is that when the `ResolveSet` bundle gets installed the `plain` provisioner will work to reconcile that content as though it were arbitrary manifests, essentially having a wrapper `BundleDeployment` for `BundleDeployments`. Generation of the `ResolveSet` bundle is trivial and can be done on the fly by the `Operator` controller. If we were to look at a `ResolveSet` bundle's manifests for the above example, it would look something like this:
```
akihikokuroda, line 118: Where is the `source` of the `ResolveSet` bundle?  Can we add the source type 'ClusterResource`?

docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
...
+```yaml
+...
+status:
+  conditions: # status conditions
+  ...
+  selections: # items selected in the form of source:name:version:channel
+  - redhat-operators:foo-package:1.2.0:platform
+  - redhat-operators:bar-package:1.1.0:platform
+  entities: # entities that were made installable via the activator-class annotation
+  - package: foo-package
```
akihikokuroda, line 110: The package is installed by Rukpak so the reference of the installable should be the contents of the `bundle.source` ?

docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
...
+status:
+  conditions: # status conditions
+  ...
+  selections: # items selected in the form of source:name:version:channel
+  - redhat-operators:foo-package:1.2.0:platform
+  - redhat-operators:bar-package:1.1.0:platform
+```
+
+Looking at the status, we can see that there is a constructed query that tells us where our resolved entitiy is located at. The `Operator` controller could read this and perform another query to get the images; However, it may be pertitnent for a new component to be introduced into Deppy - the `ResolutionActivator` (name TBD). 
+
```
akihikokuroda, line 84: Or the controller of the ActiveResolution?

docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
...
+```yaml
+...
+status:
+  conditions: # status conditions
+  ...
+  selections: # items selected in the form of source:name:version:channel
+  - redhat-operators:foo-package:1.2.0:platform
+  - redhat-operators:bar-package:1.1.0:platform
+  entities: # entities that were made installable via the activator-class annotation
```
akihikokuroda, line 109: Is this entity represented by [Entity](https://github.com/operator-framework/deppy/blob/68b8cad2c2379829cc604c0d9f6b6fadf87fbc56/internal/source/entity.go#L13)

docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
...
+
+A `ResolveSet` bundle is, at its core, just a `plain+v0` bundle of `BundleDeployments`. What this means is that when the `ResolveSet` bundle gets installed the `plain` provisioner will work to reconcile that content as though it were arbitrary manifests, essentially having a wrapper `BundleDeployment` for `BundleDeployments`. Generation of the `ResolveSet` bundle is trivial and can be done on the fly by the `Operator` controller. If we were to look at a `ResolveSet` bundle's manifests for the above example, it would look something like this:
+
+```console
+manifests/
+|- foo-package-bd.yaml
+|- bar-package-bd.yaml
+```
+
+From this, the `Operator` controller is able to maintain a single `BundleDeployment` for all of the resolved content.
```
oceanc80, line 126: May want to add a References section here where we can link the original hackmd (to capture the comments left there) and also any relevant slack threads.

tylerslaton: Updated the description to have all the context I could grab. Good call 🎉

docs/concepts/active-resolution.md
tylerslaton: So the naming here is 100% confusing to be sure but `ActiveResolution` != `ResolutionActivator`. The `ResolutionActivator` will be responsible for taking a successful resolution (in `selections`) and, using an adapter pattern, determine what is needed to make the resolution installable. It determines how to do this based on the `activator-class` annotation.

Hopefully that is coming through well, let me know where some follow-up questions may be or where I can clarify. Also totally open to rethinking this concept entirely.

docs/concepts/active-resolution.md
tylerslaton: We could update it to so that it is more consistent. This array would then look like this:

```yaml
entities: 
- id: 1234abcd
  properties:
    package: foo-package
    image: quay.io/redhat-operators/foo-package:v1.2.0
- id: 5678efgh
  properties:
    package: bar-package
    image: quay.io/redhat-operators/bar-package:v1.1.0
```

What would your thoughts be on that?

docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
+# Active Resolution
+
+The ActiveResolution API and its controller will manage the state of ActiveResolution resources. It executes the main algorithm of producing solver constraints from constraint definitions and entities, triggers resolution, converts results to something consumable by users and writes results out to the particular ActiveResolution resource.
```
timflannagan, line 3: Does the "ActiveResolution" naming imply resolution can be active or inactive? Is the intention that the "ActiveResolution" resource would be a cluster singleton, and therefore you'd have a single resource for interacting with any resolution want/needs?

Do we have an idea on whether we support multiple ActiveResolution resources in the cluster? If so, what are the implications on that approach? Is there potential for rough edges, where a resolution outcome from one AR resource influences another resolution outcome produced by another AR?

docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
...
+apiVersion: platform.openshift.io/v1alpha1
+kind: Operator
+metadata:
+  name: foo-operator
+spec:
+  package:
+    name: foo-package
+```
+
+Upon applying this resource, there are two API's that will now be at play - Deppy and RukPak. To start the process, the `Operator` controller will create an `ActiveResolution` for `Deppy` to reconcile.
```
timflannagan, line 56: What about a CreateOrUpdate-esq approach?

docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
...
+```yaml
+apiVersion: deppy.io/v1alpha1
+kind: ActiveResolution
+metadata:
+  name: foo-resolution
+spec:
+  constraints:
+  - required:
+      name: "foo-package"
+      channel: "platform"
```
timflannagan, line 67: Does this "platform" channel property get populated by the operator provisioner?

docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+For the sake of example, lets say that `foo-package` is reliant on another package called `bar-package`. In this case, `Deppy` will understand and convey this dependency to the status.
+
+```yaml
+...
+status:
+  conditions: # status conditions
+  ...
+  selections: # items selected in the form of source:name:version:channel
```
timflannagan, line 78: Note: this might be sufficient for a first pass implementation. How would this scale if the solution set to an individual resolution has high carnality? Would it be possible to hit the etcd size limitations in this scenario?

docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
...
+apiVersion: deppy.io/v1alpha1
+kind: ActiveResolution
+metadata:
+  annotations:
+    deppy.io/v1alpha1/activator-class: "Operator"
+  name: foo-resolution
+...
+```
+
+This would result in the status being updated to include a new section - entities.
```
timflannagan, line 99: What happens if someone changes the desired adapter annotation? Is the previous adapter's status.entities properties retained?

docs/concepts/active-resolution.md
akihikokuroda: I see.  So it's mapped to the `Entity`.

docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
...
+### Resolution Activator
+
+The responsibility for the `ResolutionActivator` is to be able to read `ActiveResolution`s and parse out the installable units of work. This could follow an adapter pattern, like Deppy itself does,  where the adapters are able to retrieve installable content. In the use case of `Operator`s this would result in a collections of images being made available. To mark a `ActiveResolution` with the appropriate adapter as well as if it should be activated, we can use annotations.
+
+```yaml
+apiVersion: deppy.io/v1alpha1
+kind: ActiveResolution
+metadata:
+  annotations:
+    deppy.io/v1alpha1/activator-class: "Operator"
```
timflannagan, line 94: How does this approach favor over a service who's sole job is to map the variables in the status.selections back to the defined entities by any adapter installed in the cluster? This service could either have a defined API, or an HTTP server that returns the relevant properties of an entity that higher level components could parse further?

docs/concepts/active-resolution.md
akihikokuroda: It may be out of scope right now but how it looks when the package is reference by the git source?

docs/concepts/active-resolution.md
akihikokuroda: I understood  `ActiveResolution != ResolutionActivator`.  I wonder why the controller of the ActiveResolution doesn't do the `selections` to `entities` conversion.

docs/concepts/active-resolution.md
tylerslaton: Just to make sure I understand the question - you're asking how this scales with the number of selections right?

Size limitations are definitely something we need to be aware of and I haven't thought up a great way around the limitation here. Maybe a way around this is to create a volume for each `ActiveResolution` that contains the selections and we could expose some sort of discovery API for those selections for clients?

tylerslaton: Another thought I had is that Deppy could just itself be a literal API that we hit via `https` and returns us resolutions based on cluster state instead of us needing to go through `CRD`s at all. Thoughts?

docs/concepts/active-resolution.md
tylerslaton: The `Operator API` would be responsible for creating this `ActiveResolution` resource and accordingly would set the channel that it is looking for. For a first pass I was thinking we could default this to a channel or have none. For the sake of example I included one here.

docs/concepts/active-resolution.md
```diff
@@ -0,0 +1,126 @@
...
+The ActiveResolution API and its controller will manage the state of ActiveResolution resources. It executes the main algorithm of producing solver constraints from constraint definitions and entities, triggers resolution, converts results to something consumable by users and writes results out to the particular ActiveResolution resource.
+
+## API definition
+
+Though not the final definition for this resource, the following provides an example of what it might look like:
+
+```yaml
+apiVersion: deppy.io/v1alpha1
+kind: ActiveResolution
+metadata:
```
akihikokuroda, line 12: Is this API OLM dependent?  Do we need to specify a provisioner?

docs/concepts/active-resolution.md
tylerslaton: That's a good question. In that case, at least in the current design, those status.entities would be overwritten. If we do what I described in [this comment](https://github.com/operator-framework/deppy/pull/38#discussion_r997477989) we could add some logic to that discovery API that lets you see the history of resolution.

docs/concepts/active-resolution.md
tylerslaton: It is not OLM dependent, no. The only thing that the current system may need is the `CatalogSource` API since that is utilized by one of the Deppy sources.

docs/concepts/active-resolution.md
tylerslaton: Good call, yep. I'll update the wording here to describe that to be a CreateOrUpdate interaction.

docs/concepts/active-resolution.md
akihikokuroda: I see.  I wonder it because what understands the source name.

tylerslaton:
Just a heads-up to all potential reviewers that an alternative to this approach is the framework PR that @perdasilva has been writing out. Pretty excellent work there so give it a look when you get the chance.

https://github.com/operator-framework/deppy/pull/39

PR #39, opened 2022-10-17
[WIP] Deppy as Framework Example

perdasilva:


internal/source/query.go
```diff
@@ -0,0 +1,72 @@
+package source
+
+import (
+	"sort"
+)
+
+type Predicate func(entity *Entity) bool
+
+type SearchResult []Entity
```
akihikokuroda, line 9: Is `EntityUniverse` not used here?

internal/olm/constraints.go
```diff
@@ -0,0 +1,283 @@
...
+	return o.constraints
+}
+
+func EntityToVariable(ctx context.Context, entity *source.Entity, querier source.EntityQuerier) (solver.Variable, error) {
+	solverVar := &OLMVariable{
+		id: solver.Identifier(entity.ID()),
+	}
+	constrs := make([]solver.Constraint, 0)
+
+	// collect package dependencies
```
akihikokuroda, line 45: Are the dependencies of the dependent packages added to the constraints?  I may not follow the code well.

internal/olm/constraints.go
```diff
@@ -0,0 +1,283 @@
...
+
+			resultSet, err := querier.Filter(ctx, source.And(withExportsGVK(group, version, kind)))
+			if err != nil {
+				return nil, err
+			}
+			resultSet = resultSet.Sort(byChannelAndVersion)
+			constrs = append(constrs, solver.Dependency(toSolverIdentifier(resultSet.CollectIds())...))
+		}
+	}
+
```
akihikokuroda, line 76: Are there any other (other than dependency, gvk) constraints in the entities in the channel?

hack/main.go
```diff
@@ -0,0 +1,78 @@
...
+	utilruntime.Must(v1alpha1.AddToScheme(scheme))
+}
+
+func main() {
+	/*
+		REQUIREMENTS:
+		 - fresh kind cluster
+		 - operator-sdk olm install
+		 - update /etc/hosts to point operatorhubio-catalog.olm.svc to localhost
+		 - kubectl port-forward -n olm pod/operatorhubio-catalog-xxxxx 50051:50051 (check your pod's exact name)
```
timflannagan, line 30: I feel like this could be easily scripted too, and maybe there's a Makefile target that runs that script as a prerequisite.

timflannagan, line 30: This could also work: `kubectl port-forward -n olm svc/operatorhubio-catalog 50051:50051`.

hack/main.go
```diff
@@ -0,0 +1,78 @@
...
+	err = deppySource.Sync(context.Background())
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	// create a constraint builder (or variable builder...)
+	constraintBuilder := constraints.NewConstraintBuilder(deppySource, olm.EntityToVariable,
+		constraints.WithConstraintGenerators([]constraints.ConstraintGenerator{
+			olm.RequirePackage("ack-sfn-controller", "> 1.0.0", ""),
```
timflannagan, line 51: Do I need to explicitly specify a version range here? What happens if I just want to be able to declare "I want this ack-sfn-controller package to be in the solution set, and I want deppy to find me the relevant match"?

internal/olm/source/catalog_source_deppy_source.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if apiBundle.SkipRange != "" {
+		properties["olm.skipRange"] = apiBundle.SkipRange
+	}
+
+	if len(apiBundle.Skips) > 0 {
+		properties["olm.skips"] = fmt.Sprintf("[%s]", strings.Join(apiBundle.Skips, ","))
+	}
+
```
timflannagan, line 168: I think we'd need to surface the bundle's image as well:

```go
	if apiBundle.BundlePath != "" {
		properties["olm.bundlePath"] = apiBundle.BundlePath
	}
```

Thoughts?

internal/source/query.go
perdasilva: It was before it got merged - so I just re-created it with some differences

hack/main.go
perdasilva: I mean, this is all just h4x to have the strawman. But, for sure it could be automated ^^

hack/main.go
perdasilva: yeah - I think that's reasonable. We could just check if it's "" and if so, omit the version range predicate from the entity query

perdasilva: (or change the predicate to just return true if the versionRange is empty

internal/olm/source/catalog_source_deppy_source.go
perdasilva: we can - though I think if this is to do with content, maybe it should be acquired in a different way? but I have no real preference

timflannagan: I'm not too sure what's best here. From a glance, simply translating any/all the relevant api.Bundle fields to property definitions seems reasonable to me.

internal/olm/constraints.go
```diff
@@ -0,0 +1,283 @@
...
+	v1, err := e1.GetProperty(propertyOLMVersion)
+	if err != nil {
+		return false
+	}
+	v2, err := e2.GetProperty(propertyOLMVersion)
+	if err != nil {
+		return false
+	}
+
+	return semver.MustParse(v1).GT(semver.MustParse(v2))
```
timflannagan, line 261: not blocking: handle the errors explicitly vs. relying on panics if we can't parse the incoming semver version(s)?

internal/olm/constraints.go
```diff
@@ -0,0 +1,283 @@
...
+)
+
+const (
+	propertyOLMVersion         = "olm.version"
+	propertyOLMPackageName     = "olm.packageName"
+	propertyOLMChannel         = "olm.channel"
+	propertyOLMDefaultChannel  = "olm.defaultChannel"
+	propertyOLMGVK             = "olm.gvk"
+	propertyOLMPackageRequired = "olm.package.required"
+	propertyOLMGVKRequired     = "olm.gvk.required"
```
timflannagan, line 21: Are these properties declared in any operator-registry/OLM packages that we could manually vendor here?

internal/source/source.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"context"
+)
+
+type EntityQuerier interface {
+	Get(ctx context.Context, id EntityID) *Entity
+	Filter(ctx context.Context, filter Predicate) (SearchResult, error)
+	GroupBy(ctx context.Context, fn GroupByFunction) (GroupByResult, error)
+	Iterate(ctx context.Context, fn IteratorFunction) error
```
timflannagan, line 11: not blocking: slightly surprised by the Iterate function naming given it's not returning an iterator that the call site is responsible for handling. this is fine for now though.

internal/source/source.go
```diff
@@ -0,0 +1,16 @@
+package source
+
+import (
+	"context"
+)
+
+type EntityQuerier interface {
```
timflannagan, line 7: nit: godoc?

internal/source/source.go
```diff
@@ -0,0 +1,16 @@
...
+	Get(ctx context.Context, id EntityID) *Entity
+	Filter(ctx context.Context, filter Predicate) (SearchResult, error)
+	GroupBy(ctx context.Context, fn GroupByFunction) (GroupByResult, error)
+	Iterate(ctx context.Context, fn IteratorFunction) error
+}
+
+type DeppySource interface {
```
timflannagan, line 14: nit: godoc?

internal/olm/source/catalog_source_deppy_source.go
```diff
@@ -0,0 +1,170 @@
...
+	return retry.Do(func() error {
+		var catalogSource *v1alpha1.CatalogSource = &v1alpha1.CatalogSource{}
+		if err := c.client.Get(ctx, client.ObjectKey{
+			Name:      c.name,
+			Namespace: c.namespace,
+		}, catalogSource); err != nil {
+			return err
+		}
+
+		if catalogSource.Status.GRPCConnectionState.LastObservedState != connectivity.Ready.String() {
```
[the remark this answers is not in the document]
timflannagan: Maybe this can be done at any validation layer instead when constructing a new catalogsource deppy source type?

internal/olm/source/catalog_source_deppy_source.go
```diff
@@ -0,0 +1,170 @@
...
+			return fmt.Errorf("catalog source grpc connection is not in a ready state")
```
timflannagan, line 50: nit: print the current state?

internal/olm/source/catalog_source_deppy_source.go
```diff
@@ -0,0 +1,170 @@
...
+
+func NewCatalogSourceDeppySource(kubeClient client.Client, namespace string, name string) *CatalogSourceDeppySource {
+	return &CatalogSourceDeppySource{
+		client:    kubeClient,
+		name:      name,
+		namespace: namespace,
+	}
+}
+
+func (c *CatalogSourceDeppySource) Sync(ctx context.Context) error {
```
timflannagan, line 39: Is this something we'd need to expose for client usage? The retry logic seems nice, and maybe we could allow deppy source implementations to configure it, but I wonder whether a background job could be used here that utilizes polling to periodically sync state.

go.mod
```diff
@@ -3,44 +3,52 @@ module github.com/operator-framework/deppy
...
+	github.com/avast/retry-go/v4 v4.1.0
 	github.com/go-air/gini v1.0.4
 	github.com/onsi/ginkgo/v2 v2.3.1
 	github.com/onsi/gomega v1.22.1
-	github.com/stretchr/testify v1.7.0
+	github.com/operator-framework/api v0.17.1
+	github.com/operator-framework/operator-registry v1.26.2
+	github.com/stretchr/testify v1.7.1
+	github.com/tidwall/gjson v1.14.3
+	google.golang.org/grpc v1.45.0
```
timflannagan, line 14: not blocking: this is too premature given the current state of the project, but it would be nice to avoid cluttering the repository with one-off dependencies that we'd need to maintain over time.

timflannagan:
I didn't have a ton of blocking comments, so I'm cool with merging this once the CI checks are green.

perdasilva:
closing PR as it's been broken down into a few PRs that have already been merged

PR #53, opened 2022-10-24
Deppy Framework

perdasilva:
This PR surfaces the first basic implementation of the Deppy Framework together with CLI and a couple of examples: solving dimacs files and generating (solved) sudoku boards. It makes the following changes:
- renames the `solver` package to `sat`: we may want to call it something else
- removes the `universe` package: a lot of the implementation moved to the `entitysource` package
- adds `entitysource` package: 
  - exports `Entity` implementation
  - exports `EntitySource` interface as a combination of two other interfaces: `EntityQuerier` and `EntityContentGetter`
  - exports `entitysource.Group` a struct that aggregates multiple entity sources and gives the caller a singular interface for querying and content acquisition
  - exports `CacheQuerier` an implementation of `EntityQuerier` in which all the data is in-memory
  - exports `NoContent` an implementation of `EntityContentGetter` that returns no content 
  - Note: I may have over engineered things here and separated the interfaces for querying and content acquisition
 - adds `constraint` package:
   - exports `ConstraintGenerator` and `ConstraintBuilder` interfaces
   - exports `DeppyConstraintBuilder` a simple implementation of the `ConstraintBuilder` interface
   - exports `Variable` a simple implementation of `sat.Variable` interfaces
 - adds `solver` package:
   - exports `Solver` interfaces
   - exports `DeppySolver` a simple implementation of the `Solver` interface
- adds `cmd` package: a deppy cli with two sub-commands:
 - dimacs: solves a sat problem described in dimacs format
 - sudoku: returns a random (solved) sudoku board

Notes: 
- to make dimacs work I had to change the definition of `sat.Constraint` to export the previously unexported methods (e.g. `anchor`, `apply`, etc.), so that I could define my own constraint type. I think we need to figure out a way to build these constraints with And, Or, Not top level constraints.
- sudoku seems to have a bug in it where some of the boards have gaps in them. I'm really not sure where this issue is coming from, especially since it was actually pretty easy to model the constraints using the available constraint definitions...

Co-authored-by: Catherine Chan-Tse cchantse@redhat.com


internal/sat/bench_test.go
```diff
@@ -0,0 +1,86 @@
+package sat
```
perdasilva, line 1: There are no major changes to the (old) `solver` package. I've only had to export `LitMapping` and the previously unexported `Constraint` methods, so that I could create my own constraint definition for the `dimacs` command. We should probably want to figure out a way to make the solver pluggable in some sort of way. There might be a better name for this package as well, other than `sat`, `deppysat` ?

internal/entitysource/cache_querier.go
```diff
@@ -0,0 +1,53 @@
+package entitysource
+
+import "context"
+
+var _ EntityQuerier = &CacheQuerier{}
+
+type CacheQuerier struct {
+	// TODO: separate out a cache
+	entities map[EntityID]Entity
```
perdasilva, line 9: would be good to eventually hide this behind some sort of cache interface and the method implementations just query that, or bring that cache implementation inside the CacheQuerier

internal/entitysource/entity.go
```diff
@@ -0,0 +1,35 @@
+package entitysource
+
+import "fmt"
+
+type EntityID string
+
+type EntityPropertyNotFoundError string
```
perdasilva, line 7: Do we want this? or should we just use regular go mechanics, `e.g. value, ok := m[key]`?

internal/entitysource/entity_source.go
```diff
@@ -0,0 +1,88 @@
...
+	Filter(ctx context.Context, filter Predicate) (SearchResult, error)
+	GroupBy(ctx context.Context, fn GroupByFunction) (GroupByResult, error)
+	Iterate(ctx context.Context, fn IteratorFunction) error
+}
+
+type EntityContentGetter interface {
+	GetContent(ctx context.Context, id EntityID) (interface{}, error)
+}
+
+type EntitySource interface {
```
perdasilva, line 18: over engineering?

cmd/sudoku/sudoku.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func NewSudoku() *Sudoku {
+	var entities = make(map[entitysource.EntityID]entitysource.Entity, 9*9*9)
+	for row := 0; row < 9; row++ {
+		for col := 0; col < 9; col++ {
+			for num := 0; num < 9; num++ {
+				id := GetId(row, col, num)
+				entities[id] = *entitysource.NewEntity(id, map[string]string{
+					"row": strconv.Itoa(row),
```
perdasilva, line 36: these properties are kind of superfluous since I don't do any querying outside getting by Id

internal/entitysource/entity_source.go
```diff
@@ -0,0 +1,88 @@
...
+type EntityQuerier interface {
+	Get(ctx context.Context, id EntityID) *Entity
+	Filter(ctx context.Context, filter Predicate) (SearchResult, error)
+	GroupBy(ctx context.Context, fn GroupByFunction) (GroupByResult, error)
+	Iterate(ctx context.Context, fn IteratorFunction) error
+}
+
+type EntityContentGetter interface {
+	GetContent(ctx context.Context, id EntityID) (interface{}, error)
+}
```
akihikokuroda, line 16: Does the each type of the source implement `EntityContentCetter` interface?

internal/entitysource/search.go
```diff
@@ -1,15 +1,10 @@
...
 
-	"github.com/operator-framework/deppy/internal/source"
-)
+type Predicate func(entity *Entity) bool
 
-type Predicate func(entity *source.Entity) bool
-
-type SearchResult []source.Entity
-type SortFunction func(e1 *source.Entity, e2 *source.Entity) bool
+type SearchResult []Entity
```
akihikokuroda, line 7: I don't like the types named xxxResult.  They holds the results of a specific function call but I like some names that represent the contents type.  so like

EntitySet instead of SerchResult and GroupedEntitySet instead of GroupByResult.  WDYT?

perdasilva: I like that! I'll make the changes

cmd/sudoku/sudoku.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"github.com/operator-framework/deppy/internal/constraints"
+	"github.com/operator-framework/deppy/internal/entitysource"
+	"github.com/operator-framework/deppy/internal/sat"
+)
+
+var _ entitysource.EntitySource = &Sudoku{}
+var _ constraints.ConstraintGenerator = &Sudoku{}
+
+type Sudoku struct {
```
akihikokuroda, line 18: Can we add
```
	*entitysource.CacheQuerier
	entitysource.NoContentSource
```

and take out 

```
func (s Sudoku) Get(_ context.Context, id entitysource.EntityID) *entitysource.Entity {
func (s Sudoku) Filter(ctx context.Context, filter entitysource.Predicate) (entitysource.SearchResult, error) {
func (s Sudoku) GroupBy(ctx context.Context, fn entitysource.GroupByFunction) (entitysource.GroupByResult, error) {
func (s Sudoku) Iterate(ctx context.Context, fn entitysource.IteratorFunction) error {
func (s Sudoku) GetContent(ctx context.Context, id entitysource.EntityID) (interface{}, error) {
```
?

internal/entitysource/entity_source.go
```diff
@@ -0,0 +1,88 @@
...
+
+type EntityQuerier interface {
+	Get(ctx context.Context, id EntityID) *Entity
+	Filter(ctx context.Context, filter Predicate) (SearchResult, error)
+	GroupBy(ctx context.Context, fn GroupByFunction) (GroupByResult, error)
+	Iterate(ctx context.Context, fn IteratorFunction) error
+}
+
+type EntityContentGetter interface {
+	GetContent(ctx context.Context, id EntityID) (interface{}, error)
```
akihikokuroda, line 15: Does this retrieve just one entity from the source?

timflannagan (commented):
Would it be possible to separate this overall PR into smaller PRs that built on top of each other?

internal/entitysource/entity_source.go
perdasilva: I just thought separating these concerns made some sense - but, idk, could be over engineering

cmd/sudoku/sudoku.go
perdasilva: Great catch! Fixing it!

internal/entitysource/entity_source.go
perdasilva: This retrieved any content associated with the id. This is a bit more forward looking for the OLM use case, we'll do the calculation over bundle metadata, but eventually will also need to get the bundles themselves for installation

internal/constraints/constraint_builder.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/operator-framework/deppy/internal/entitysource"
+	"github.com/operator-framework/deppy/internal/sat"
+)
+
+type ConstraintGenerator interface {
+	GetVariables(ctx context.Context, querier entitysource.EntityQuerier) ([]sat.Variable, error)
+}
+
+type ConstraintBuilder interface {
```
akihikokuroda, line 14: This is the exact same interface as the `ConstraintGenerator`.   Do we need both?

perdasilva:
> Would it be possible to separate this overall PR into smaller PRs that built on top of each other?

@timflannagan can do, I'll just address the last comments

internal/constraints/constraint_builder.go
perdasilva: Good point - I've removed the ConstraintBuilder

perdasilva:
closing this PR as I've broken it down into smaller PRs for easy of review

PR #55, opened 2022-10-25
OLM constraints

perdasilva:
This PR adds the constraint generators for the basic OLM constraints: package required, gvk uniqueness, package uniqueness, pkg dependency and gvk dependency.


internal/ext/olm/constraints_test.go
```diff
@@ -0,0 +1,542 @@
...
+			})
+			It("finds no candidates to satisfy the dependency when no entries contain the 'olm.packageName' key", func() {
+				mockQuerier.testEntityList = entitysource.EntityList{
+					*entitysource.NewEntity("cool-package-3-entity", map[string]string{
+						"wrong-key":        "cool-package-1",
+						propertyOLMVersion: "2.1.2",
+						propertyOLMChannel: "channel-1",
+					}),
+				}
+				satVars, err := olm.RequirePackage("cool-package-4", "<=3.0.0", "channel-3").GetVariables(ctx, mockQuerier)
```
akihikokuroda, line 147: Should the package name be "cool-package-1" instead of "cool-package-4"?

dtfranz: Yes I think you're right. The correct key is missing on the entity that gets returned so they wouldn't match anyway but it's probably important in the test to show that it's the missing key and not the mis-matched package names that's giving the expected test result.

internal/ext/olm/constraints_test.go
```diff
@@ -0,0 +1,542 @@
...
+				satVars, err := olm.PackageUniqueness().GetVariables(ctx, mockQuerier)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(satVars).Should(HaveLen(3))
+				sort.Slice(satVars, func(i, j int) bool {
+					return satVars[i].Identifier().String() < satVars[j].Identifier().String()
+				})
+				Expect(satVars[0].Identifier().String()).To(Equal("cool-package-1 uniqueness"))
+				Expect(satVars[0].Constraints()[0].String("test-pkg")).To(Equal("test-pkg permits at most 1 of cool-package-1-entity"))
+				Expect(satVars[1].Identifier().String()).To(Equal("cool-package-2 uniqueness"))
+				Expect(satVars[1].Constraints()[0].String("test-pkg")).To(Equal("test-pkg permits at most 1 of cool-package-2-1-entity, cool-package-2-0-entity"))
```
akihikokuroda, line 272: The order if of the packages `cool-package-2-1-entity, cool-package-2-0-entity` should be flexible in the test.

internal/ext/olm/constraints_test.go
```diff
@@ -0,0 +1,542 @@
...
+			})
+			It("returns a slice of sat.Variable grouped by group, version, and kind, with constraints ordered by package name", func() {
+				satVars, err := olm.GVKUniqueness().GetVariables(ctx, mockQuerier)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(satVars).Should(HaveLen(2))
+				sort.Slice(satVars, func(i, j int) bool {
+					return satVars[i].Identifier().String() < satVars[j].Identifier().String()
+				})
+				Expect(satVars[0].Identifier().String()).To(Equal("my-group/my-version/my-kind uniqueness"))
+				Expect(satVars[0].Constraints()[0].String("foo")).To(Equal("foo permits at most 1 of cool-package-1-entity, cool-package-2-0-entity, cool-package-3-entity"))
```
akihikokuroda, line 316: The order if of the packages `cool-package-1-entity, cool-package-2-0-entity, cool-package-3-entity` should be flexible in the test.

internal/ext/olm/constraints_test.go
```diff
@@ -0,0 +1,542 @@
...
+					*entitysource.NewEntity("cool-package-3-entity", map[string]string{
+						propertyOLMPackageName: "cool-package-3",
+						propertyOLMGVK:         "abcdefg",
+					}),
+				}
+				satVars, err := olm.GVKUniqueness().GetVariables(ctx, mockQuerier)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(satVars).Should(HaveLen(0))
+			})
+			It("does not panic but returns a constraint with '//' id when gvk json is missing fields", func() {
```
akihikokuroda, line 349: This is interesting test.  Should it return an entry in satVars for `//`?  Does it constraint uniqueness of the packages with `{}` in the `propertyOLMGVK:`?  I think that `{}` should be ignored in the runtime code.

internal/ext/olm/constraints.go
```diff
@@ -0,0 +1,298 @@
...
+					return true
+				}
+			}
+		}
+		return false
+	}
+}
+
+func byChannelAndVersion(e1 *entitysource.Entity, e2 *entitysource.Entity) bool {
+	e1pkgName, err := e1.GetProperty(propertyOLMPackageName)
```
akihikokuroda, line 205: When the package name of both e1 and e2 are not defined, the return value of this method is the same for `byChannedAndVersion(e1, e2)` and `byChannedAndVersion(e1, e2)`.   So the result of sorting depends on the order of the fed entities.  It may not be good.

akihikokuroda (commented):
Looks good for me.  I have some minor comments.  Thanks

internal/ext/olm/constraints_test.go
```diff
@@ -0,0 +1,542 @@
...
+				Expect(satVars[0].Constraints()[0].String("test-pkg")).To(Equal("test-pkg is mandatory"))
+				Expect(satVars[0].Constraints()[1].String("test-pkg")).To(Equal("test-pkg requires at least one of cool-package-1-entity"))
+			})
+			// package name
+			It("finds no candidates to satisfy the dependency when package name does not match any entities", func() {
+				satVars, err := olm.RequirePackage("cool-package-4", "<3.0.0", "channel-1").GetVariables(ctx, mockQuerier)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(satVars).Should(HaveLen(1))
+				Expect(satVars[0].Constraints()).Should(HaveLen(2))
+				Expect(satVars[0].Constraints()[1].String("test-pkg")).To(Equal("test-pkg has a dependency without any candidates to satisfy it"))
```
awgreene, line 137: Where is the "test-pkg" constraint introduced?

internal/ext/olm/constraints_test.go
```diff
@@ -0,0 +1,542 @@
...
+						propertyOLMChannel:     "",
+					}),
+				}
+				satVars, err := olm.RequirePackage("cool-package-1", "<=3.0.0", "channel-3").GetVariables(ctx, mockQuerier)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(satVars).Should(HaveLen(1))
+				Expect(satVars[0].Constraints()).Should(HaveLen(2))
+				Expect(satVars[0].Constraints()[1].String("test-pkg")).To(Equal("test-pkg has a dependency without any candidates to satisfy it"))
+			})
+			It("returns one satVar entry describing dependency candidate when channel requirement is empty", func() {
```
awgreene, line 211:
```suggestion
			It("returns one satVar entry describing no candidate when channel requirement is empty", func() {
```

internal/ext/olm/constraints_test.go
```diff
@@ -0,0 +1,542 @@
...
+						propertyOLMChannel:     "channel-1",
+					}),
+				}
+				satVars, err := olm.RequirePackage("cool-package-1", "<=3.0.0", "channel-3").GetVariables(ctx, mockQuerier)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(satVars).Should(HaveLen(1))
+				Expect(satVars[0].Constraints()).Should(HaveLen(2))
+				Expect(satVars[0].Constraints()[1].String("test-pkg")).To(Equal("test-pkg has a dependency without any candidates to satisfy it"))
+			})
+			// channel
```
awgreene, line 196: Should a successful case be tested?

internal/ext/olm/constraints_test.go
awgreene: Yeah I'm confused here as well. What is the user trying to specify with an empty GVK property? Why are we returning any result?

internal/ext/olm/constraints_test.go
```diff
@@ -0,0 +1,542 @@
...
+				}),
+			}
+			satVars, err := olm.GVKUniqueness().GetVariables(ctx, mockQuerier)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(satVars).Should(HaveLen(1))
+			entities := strings.Split(satVars[0].Constraints()[0].String("pkg"), ", ")
+			Expect(entities).To(MatchAllElementsWithIndex(IndexIdentity, matchElements))
+		},
+			Entry("orders by packageName when both keys exist", propertyOLMPackageName, propertyOLMPackageName, Elements{
+				"0": Equal("pkg permits at most 1 of cool-package-entity-1"),
```
awgreene, line 456: Why does this read as `pkg permits at most 1 of cool-package-entity-1, cool-package-entity-2` as other constraints do? It isn't clear to me why `cool-package-entity-2` is stored separately in the array.

internal/ext/olm/constraints_test.go
```diff
@@ -0,0 +1,542 @@
...
+			}),
+		)
+		Describe("channel and version ordering", func() {
+			It("orders sat vars with identical packageName by channel and version in that order of priority", func() {
+				mockQuerier.testEntityList = entitysource.EntityList{
+					*entitysource.NewEntity("cool-package-1-ch1-1.0-entity", map[string]string{
+						propertyOLMPackageName:    "cool-package-1",
+						propertyOLMVersion:        "1.0.1",
+						propertyOLMChannel:        "channel-1",
+						propertyOLMDefaultChannel: "channel-1",
```
awgreene, line 475: Why do entities, which I believe represent bundles, have a default channel property?

awgreene (commented):
Looks great, I had a bunch of questions regarding stuff I should have learned awhile ago :)

internal/ext/olm/constraints_test.go
akihikokuroda: The `String` function of the `sat.Variabe` takes a string argument.  The subject name (or ID) of the constraint is usually provided there but the `String` function just embedded it in the message.

internal/ext/olm/constraints_test.go
perdasilva: this is important because you need to sort the bundles in channel order with the default channel at the front - in case no channel is given

internal/ext/olm/constraints_test.go
perdasilva: fixed

internal/ext/olm/constraints_test.go
perdasilva: nice catch! I've fixed it up. We still won't get an error but the result set will be zero.

internal/ext/olm/constraints.go
perdasilva: fair call - I'll change it to sort by id

perdasilva: I've updated the sorting function to push unknown values to the bottom of the pile. If we get down to comparing version then it will be by entity id

internal/ext/olm/constraints.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/operator-framework/deppy/internal/sat"
+)
+
+const (
+	propertyOLMGVK            = "olm.gvk"
+	propertyOLMPackageName    = "olm.packageName"
+	propertyOLMVersion        = "olm.version"
+	propertyOLMChannel        = "olm.channel"
+	propertyOLMDefaultChannel = "olm.defaultChannel"
+)
```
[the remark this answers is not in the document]
perdasilva: Atm, Deppy doesn't care about properties. I.e. there's nothing in the framework at the moment that can see a property of a particular type and it knows how to handle them. Properties are just something the entities come with. If on the source side new properties are added to the entities, something on the constraint generation side will need to know how to interpret them.

internal/ext/olm/constraints.go
```diff
@@ -0,0 +1,298 @@
...
+func (r *requirePackage) GetVariables(ctx context.Context, querier entitysource.EntityQuerier) ([]sat.Variable, error) {
+	resultSet, err := querier.Filter(ctx, entitysource.And(
+		withPackageName(r.packageName),
+		withinVersion(r.versionRange),
+		withChannel(r.channel)))
+	if err != nil {
+		return nil, err
+	}
+	ids := resultSet.Sort(byChannelAndVersion).CollectIds()
+	subject := subject("require", r.packageName, r.versionRange, r.channel)
```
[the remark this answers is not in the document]
perdasilva: It can be, but I don't think it has to be since the scope of this string is only this method. This is just used for generating the variable ID, which in general ought to be unique. Outside of providing a user readable intent, and being unique, it has no use, e.g., programmatically (at least not at the moment)

internal/ext/olm/constraints.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil {
+		return nil, err
+	}
+	variables := make([]sat.Variable, 0, len(resultSet))
+	for key, entities := range resultSet {
+		ids := toSatIdentifier(entities.Sort(byChannelAndVersion).CollectIds()...)
+		variables = append(variables, constraints.NewVariable(u.subject(key), sat.AtMost(1, ids...)))
+	}
+	return variables, nil
+}
```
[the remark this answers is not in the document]
perdasilva: I'll add a comment for clarity, though documentation will be tracked in another ticket (we def. need some tho!).

So, the current idea of the framework is the following:
1. EntitySource provides a queryable interface for acquiring entities that have certain attributes. Thing like, give me every entity where property foo == bar, group entities by property foo -> {"bar": [e1, e2], "fiz": [e3, e4], "buz": [e1, e5, e6], ...}
2. ConstraintGenerators are given an entitysource interface, which they use to get the entities that they care about and create constraints
3. Solver takes entity sources and constraint generators and weaves them: given these entity sources, and these constraint generators, give me the constraints I should feed to my solver to get a solution.

The uniqueness constraints state that you should only pick at most one bundle per package and at most one bundle per gvk. It uses a groupBy query to get all bundles / package and all bundle / gvk, then creates the necessary variable with constraints: for each pkg: atMost(1, pkg.bundles); for each gvk: atMost(1, gvk.bundles).

Another example. For required packages:
Constraint generator says: entitysource, please give me entities with package property `p`, channel property `c`, version property in version range `r`: then outputs a variable with a `mandatory` and dependency (on the resulting entities) constraints.

internal/ext/olm/constraints.go
```diff
@@ -0,0 +1,298 @@
...
+func PackageUniqueness() constraints.ConstraintGenerator {
+	return &uniqueness{
+		subject:   uniquenessSubjectFormat,
+		groupByFn: packageGroupFunction,
+	}
+}
+
+func uniquenessSubjectFormat(key string) sat.Identifier {
+	return sat.IdentifierFromString(fmt.Sprintf("%s uniqueness", key))
+}
```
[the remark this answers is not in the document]
perdasilva: what would be more descriptive? `... bundle uniqueness`? Or, more verbiage like `only one bundle from gvk (...) may be selected` and `only one bundle from package (...) may be selected`?

internal/ext/olm/constraints.go
```diff
@@ -0,0 +1,298 @@
...
+	if gvks, err := entity.GetProperty(propertyOLMGVK); err == nil {
+		gvkArray := gjson.Parse(gvks).Array()
+		keys := make([]string, len(gvkArray))
+		for i, val := range gvkArray {
+			gvk := fmt.Sprintf("%s/%s/%s", val.Get("group"), val.Get("version"), val.Get("kind"))
+			keys[i] = gvk
+		}
+		return keys
+	}
+	return nil
```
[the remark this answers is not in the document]
perdasilva: this would be called once per resolution - have to keep in mind that the data behind the entity source can change in the meantime. I'm kicking around the idea of using generics in the framework. This could potentially improve performance at the expense of genericity. E.g. the entities you kick around would be OLMBundleEntities. But then, the constraint generators would also have to use OLMBundleEntities, which would make them less generic and applicable to other solvers.

internal/ext/olm/constraints_test.go
```diff
@@ -0,0 +1,542 @@
...
+	RunSpecs(t, "Constraints Suite")
+}
+
+const (
+	propertyOLMGVK            = "olm.gvk"
+	propertyOLMPackageName    = "olm.packageName"
+	propertyOLMVersion        = "olm.version"
+	propertyOLMChannel        = "olm.channel"
+	propertyOLMDefaultChannel = "olm.defaultChannel"
+)
```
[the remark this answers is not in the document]
perdasilva: I think this is a side effect of the separation between the querying and constraint generation layers. I guess my first instinct was to put them down as an internal implementation detail. But it might be worth exporting them in case people want to get particular property values out of the entities.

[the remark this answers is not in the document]
perdasilva: I've changed it to expose the properties and updated the test

internal/ext/olm/constraints_test.go
perdasilva: Exactly, it's an artifact of the sat.Constraint interface

internal/ext/olm/constraints_test.go
perdasilva: Actually, the problem here is there is a lack of checking whether the filter query actually produced any results. Rn, it will return a variable with a dependency on nothing. I'll fix.

internal/ext/olm/constraints_test.go
perdasilva: Fixed

internal/ext/olm/constraints_test.go
perdasilva: This is also an artifact of the sat.Constraints. There's currently no way to get the dependency ids. What was done here in lieu of that is to parse the string representation of the constraint. We should probably rethink the sat.Constraint interface

akihikokuroda (approved):
LGTM.  Thanks!

internal/ext/olm/constraints.go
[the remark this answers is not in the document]
awgreene: My original concern was that `my-group/my-version/my-kind uniqueness` only implies GVK uniqueness if you know that this `my-group/my-version/my-kind` is a GVK. The same applies to `cool-package-1 uniqueness`. Not a blocker, but something we may want to consider.

internal/ext/olm/constraints_test.go
awgreene: SGTM, not a blocker. We'll probably need an entire sprint to make sat results readable :)

internal/ext/olm/constraints_test.go
awgreene: Got it, thanks @perdasilva!

internal/ext/olm/constraints_test.go
awgreene: Thanks @akihikokuroda

PR #56, opened 2022-10-25
Bump the golangci-lint tool dependency to v1.49.0

timflannagan:
It looks like we already had the Makefile scaffoldings but golangci-lint wasn't being run in the CI pipeline. Adds the .golanci-lint.yaml configuration file, and leverages the existing "lint" and "verify" Makefile targets to be runs as GHA jobs.

### DIFF
--- a/.github/workflows/sanity.yaml
+++ b/.github/workflows/sanity.yaml
@@ -22,6 +22,9 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: "go.mod"
 
       - name: Cache build and module paths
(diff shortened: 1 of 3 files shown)


.github/workflows/sanity.yaml
```diff
@@ -22,6 +22,9 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: "go.mod"
```
timflannagan, line 27: The more recent golangci-lint minor versions require 1.18 or 1.19 Go versions, and it looks like we're pulling in the default Go version given we weren't explicitly setting up the Go version before.

exdx (approved):

PR #57, opened 2022-10-25
Deppy Framework (1/7): reset and rename solver package to sat

perdasilva:
This PR:
* renames the solver package to sat: we may want to call it something else
* removes the universe package: a lot of the implementation will move to the entitysource package
* opens up the sat `Constraint` interface to be implemented outside of the `sat` package and exports LitMapping


timflannagan (approved):
I didn't fully understand the changes from order -> Order, and apply -> Apply, but I don't think that's blocking and this is fine for now.

oceanc80 (approved):

akihikokuroda (approved):

PR #58, opened 2022-10-25
Deppy Framework (2/7): EntitySource API

perdasilva:
This PR:

* adds entitysource package:
* exports Entity implementation
* exports EntitySource interface as a combination of two other interfaces: EntityQuerier and EntityContentGetter
* exports entitysource.Group a struct that aggregates multiple entity sources and gives the caller a singular interface for querying and content acquisition
* exports CacheQuerier an implementation of EntityQuerier in which all the data is in-memory
* exports NoContent an implementation of EntityContentGetter that returns no content
**Note**: I may have over engineered things here and separated the interfaces for querying and content acquisition

### DIFF
--- a/internal/entitysource/cache_querier.go
+++ b/internal/entitysource/cache_querier.go
@@ -1,0 +1,53 @@
+package entitysource
+
+import "context"
+
+var _ EntityQuerier = &CacheQuerier{}
+
+type CacheQuerier struct {
+	// TODO: separate out a cache
+	entities map[EntityID]Entity
+}
+
+func NewCacheQuerier(entities map[EntityID]Entity) *CacheQuerier {
+	return &CacheQuerier{
+		entities: entities,
+	}
+}
+
+func (c CacheQuerier) Get(_ context.Context, id EntityID) *Entity {
+	if entity, ok := c.entities[id]; ok {
+		return &entity
+	}
+	return nil
+}
+
+func (c CacheQuerier) Filter(_ context.Context, filter Predicate) (EntityList, error) {
+	resultSet := EntityList{}
+	for _, entity := range c.entities {
+		if filter(&entity) {
+			resultSet = append(resultSet, entity)
+		}
+	}
+	return resultSet, nil
+}
+
+func (c CacheQuerier) GroupBy(_ context.Context, fn GroupByFunction) (EntityListMap, error) {
+	resultSet := EntityListMap{}
+	for _, entity := range c.entities {
+		keys := fn(&entity)
+		for _, key := range keys {
+			resultSet[key] = append(resultSet[key], entity)
+		}
+	}
+	return resultSet, nil
+}
+
+func (c CacheQuerier) Iterate(_ context.Context, fn IteratorFunction) error {
+	for _, entity := range c.entities {
+		if err := fn(&entity); err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/internal/entitysource/entity_source.go
+++ b/internal/entitysource/entity_source.go
@@ -1,0 +1,110 @@
+package entitysource
+
+import (
+	"context"
+)
+
+// IteratorFunction is executed for each entity when iterating over all entities
+type IteratorFunction func(entity *Entity) error
+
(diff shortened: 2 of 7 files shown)


internal/entitysource/cache_querier.go
```diff
@@ -0,0 +1,53 @@
+package entitysource
+
+import "context"
+
+var _ EntityQuerier = &CacheQuerier{}
+
+type CacheQuerier struct {
```
akihikokuroda, line 7: Is the way to update the cache missing?  Should it be recreated every time?

internal/entitysource/entity_source.go
```diff
@@ -0,0 +1,88 @@
...
+
+type EntityQuerier interface {
+	Get(ctx context.Context, id EntityID) *Entity
+	Filter(ctx context.Context, filter Predicate) (SearchResult, error)
+	GroupBy(ctx context.Context, fn GroupByFunction) (GroupByResult, error)
+	Iterate(ctx context.Context, fn IteratorFunction) error
+}
+
+type EntityContentGetter interface {
+	GetContent(ctx context.Context, id EntityID) (interface{}, error)
```
akihikokuroda, line 15: Does `GetContent" retrieve the entity from the source?  Is the retrieve entity put in the the cache of the EntityQuerier?

internal/entitysource/cache_querier.go
perdasilva: I think we should hide this entity list behind some caching interface that handles that sort of thing. For now the `cache` aspect is just that it's in memory

internal/entitysource/entity_source.go
perdasilva: No, content is meant to be linked to some blob (e.g. CatalogSource bundle) that is linked to the EntityID. 
The idea here is, you use querier for building constraints, then once you have the IDs you care about, you can pull the content for them (e.g. after resolving the bundles, you need to install them)

internal/entitysource/search.go
```diff
@@ -0,0 +1,70 @@
+package entitysource
```
akihikokuroda, line 1: Why is this file named `search.go`?  There is no search functions in it.

internal/entitysource/search.go
```diff
@@ -0,0 +1,70 @@
...
+	ids := make([]EntityID, len(r))
+	for i := range r {
+		ids[i] = r[i].ID()
+	}
+	return ids
+}
+
+type SortFunction func(e1 *Entity, e2 *Entity) bool
+
+type IteratorFunction func(entity *Entity) error
```
akihikokuroda, line 26: Would you move some of the type definitions and Predicate function definitions to entity_source.go or cache_querier.go because they are not used in this file?

internal/entitysource/cache_querier.go
akihikokuroda: I see.  Are the all entity source adapter(?) implement this interface?  If the list (cache) is here all functions in this interface (struct) have the same implementation.  https://github.com/operator-framework/deppy/pull/52

akihikokuroda:
Can we have unit tests for entity_source and cache_querier?

internal/entitysource/cache_querier.go
perdasilva: They can, but they don't have to. This is just meant as a simple in-memory implementation. We might want to revisit the querying interface as it may not be easily generalizable. But, to start with and prove the flow, it should be ok.

internal/entitysource/search.go
perdasilva: I'm changing it to `query.go` would that be better?

perdasilva:
> Can we have unit tests for entity_source and cache_querier?

We definitely need those. I think we are trying to get something up asap, so we can iterate over it. So, the unit tests should come in a follow-up PR

internal/entitysource/search.go
perdasilva: Done

akihikokuroda (approved):
LGTM Thanks!

PR #60, opened 2022-10-25
Deppy Framework (4/7): Solver API

perdasilva:
This PR adds the `solver` package:
 * exports Solver interface
 * exports DeppySolver a simple implementation of the Solver interface


internal/solver/solver.go
```diff
@@ -0,0 +1,72 @@
...
+
+var _ Solver = &DeppySolver{}
+
+type Solution map[entitysource.EntityID]bool
+
+type Solver interface {
+	Solve(ctx context.Context) (Solution, error)
+}
+
+type Options struct {
```
akihikokuroda, line 20: Isn't it easier to put these items separately as the arguments for NewDepplySolver instead of defining a structure and putting it as an argument?

internal/solver/solver.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+
+	"github.com/operator-framework/deppy/internal/constraints"
+	"github.com/operator-framework/deppy/internal/entitysource"
+	"github.com/operator-framework/deppy/internal/sat"
+)
+
+var _ Solver = &DeppySolver{}
+
+type Solution map[entitysource.EntityID]bool
```
timflannagan, line 14: nit: godoc if we're exporting this and expecting clients to handle this output in some capacity?

internal/solver/solver.go
timflannagan: Yeah, same question as Aki here. Is the intention to have a config-esq structure here? I'd imagine a caller creating a NewDeppySolver would like to avoid having to create an auxillary type here and instead relying on the default configuration. If they want to extend the default configuration, then we could leverage the options pattern here?

timflannagan (approved):
Nice, this seems fine to me for an initial implementation. Had some questions around the Options structure, but I don't think those are blocking.

internal/solver/solver.go
```diff
@@ -0,0 +1,72 @@
...
+		if entity := querier.Get(ctx, entitysource.EntityID(variable.Identifier())); entity != nil {
+			solution[entity.ID()] = false
+		}
+	}
+	for _, variable := range selection {
+		if entity := querier.Get(ctx, entitysource.EntityID(variable.Identifier())); entity != nil {
+			solution[entity.ID()] = true
+		}
+	}
+	return solution, nil
```
timflannagan, line 71: Do we have an idea on how callers would be utilizing this information? I think we can add tests after this merges so we have a barebones framework implementation, but it would be nice to figure out whether this is flexible enough for client usage.

perdasilva: To be honest, this was a knee-jerk reaction to the two commands I put it. Sometimes we only care about what was selected, others, you might care about the whole thing. I figured this was the easiest way to start with. We could also have it as some kind of interface where you query the assignment for a variable and can get all variables that were considered.

timflannagan: Create a follow-up ticket for re-evaluating the return value then?

internal/solver/solver.go
perdasilva: fixed

internal/solver/solver.go
```diff
@@ -0,0 +1,63 @@
...
+// Solution is returned by the Solver when a solution could be found
+// it can be queried by EntityID to see if the entity was selected (true) or not (false)
+// by the solver
+type Solution map[entitysource.EntityID]bool
+
+type Solver interface {
+	Solve(ctx context.Context) (Solution, error)
+}
+
+// DeppySolver is a simple solver implementation that takes a
```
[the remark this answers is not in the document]
perdasilva: ugh - my bad

oceanc80, line 22: Missing the end of the godoc :)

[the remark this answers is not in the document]
perdasilva: too many things at once, got distracted. I've fixed it =D

perdasilva: fixed ^^

PR #61, opened 2022-10-25
Deppy Framework (5/7): root deppy command

perdasilva:
This PR adds the root deppy command and associated make target

### DIFF
--- a/cmd/root/root.go
+++ b/cmd/root/root.go
@@ -1,0 +1,14 @@
+package root
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func NewRootCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "deppy",
+		Short: "Deppy is an open-source constraint solver framework",
+		Long: `An open-source constraint solver framework written in Go.
+For more information visit https://github.com/operator-framework/deppy`,
+	}
+}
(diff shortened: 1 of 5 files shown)


oceanc80 (approved):

cmd/root/root.go
```diff
@@ -0,0 +1,22 @@
...
+var rootCmd = &cobra.Command{
+	Use:   "deppy",
+	Short: "Deppy is an open-source constraint solver framework",
+	Long: `An open-source constraint solver framework written in Go.
+For more information visit https://github.com/operator-framework/deppy`,
+}
+
+func Execute() {
+	if err := rootCmd.Execute(); err != nil {
+		_, _ = fmt.Fprintln(os.Stderr, err)
```
timflannagan, line 19: Any reason why we can't just do `fmt.Fprintln(os.Stderr, err)` here without capturing any of the return values?

cmd/root/root.go
```diff
@@ -0,0 +1,22 @@
+package root
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var rootCmd = &cobra.Command{
```
timflannagan, line 10: nit: Use a constructor that returns a *cobra.Command vs. a packge-level variable?

cmd/root/root.go
perdasilva: I just get annoyed by the squiggles in the ide...this was just to print to stderr btw

cmd/root/root.go
perdasilva: I've done it. But, how come?

joelanford: generally in my experience, globals variables are headaches

PR #67, opened 2022-10-26
Rename `internal` package to `pkg` to make framework accessible

perdasilva:


akihikokuroda:
I feel this is too much.  I hope we can selectively choose and separate the interfaces to expose

perdasilva:
> I feel this is too much. I hope we can selectively choose and separate the interfaces to expose

That's a fair call. I'll close the PR and maybe we can take a more piecemeal approach ^^

PR #71, opened 2022-11-04
[WIP] deppy catsrc adapter

ankitathomas:
Implementation of a deppy source adapter that can provide a ListEntities functionality.

This prototype uses a single CatalogSource specified either by its namespaced name or its serving address as a data source. This model necessitates the addition of another controller layer on top of the adapter to create and monitor adapters for all catalogsources and lacks any caching at the adapter side, which may be performance concerns. Some of these issues may be solvable by a cluster-scoped unified data source. 

See https://github.com/anik120/rukpak-packageserver


internal/source/adapter/api/deppy_source_adapter.proto
```diff
@@ -0,0 +1,24 @@
...
+message EntityID {
+  string source = 1;
+  string package = 2;
+  string version = 3;
+  string name = 4;
+}
+
+message Entity {
+  EntityID id = 1;
+  map<string,string> properties = 2;
```
joelanford, line 22: (I know this is not in the design doc, so this is a design quesiton, not an impl question)...

Do we need a separate `constraints` field here?

- Properties that are not understood can likely be ignored by deppy
- Constraints that are not understood should probably result in a resolution error (or at least there should be a knob that lets the user decide whether unknown constraints should be failures)

Without a separate constraints field, it seems like deppy will need to have knowledge about how to extract constraints from properties.

internal/source/adapter/api/deppy_source_adapter.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
ankitathomas: I looked into OpenAPI as an alternative for an HTTP/JSON endpoint, but it didn't make the building out of the API any easier or its definition simpler. There are other options, but GRPC seemed more straightforward for an initial API.

internal/source/adapter/api/deppy_source_adapter.proto
ankitathomas: Constraints should be something that the layer above the adapter understands, it didn't seem appropriate to assume constraints directly from the catalogsource.

ankitathomas:
Closing in favor of https://github.com/operator-framework/operator-controller/pull/129

PR #79, opened 2022-12-15
move internal to pkg

perdasilva:
Moves internal to pkg to facilitate importing deppy into other project


akihikokuroda (approved):
I have a little concern about the sat package exposure.  I hope it will be refactored before it become too late.

PR #81, opened 2022-12-20
remove entitysource group and variableaggregator

perdasilva:
This PR simplifies the framework slightly by:
* Removing the concept of EntitySource Group
* Removing the concept of ConstraintAggregators
* Collapsing the EntitySource interface to _just_ the query methods

Reasoning:
EntitySource Groups and Constraint Aggregators are nothing more than implementations of the EntitySource and ConstraintGenerators (now called VariableSources). They don't need to live in the framework, at least not for now.

I've removed the GetContent method from the EntitySource interface. Not all entities have content backing them, for some that do, the content can be carried in the EntityProperties (e.g. image path, etc.). For those that _do_ have content, a custom EntitySource that adds the required method(s) can be created by the user.


pkg/entitysource/entity_source.go
```diff
@@ -20,91 +20,10 @@ type Predicate func(entity *Entity) bool
...
-// EntityQuerier is an interface for querying entities in some store
-type EntityQuerier interface {
+// EntitySource provides a query and content acquisition interface for arbitrary entity stores
+type EntitySource interface {
 	Get(ctx context.Context, id EntityID) *Entity
 	Filter(ctx context.Context, filter Predicate) (EntityList, error)
 	GroupBy(ctx context.Context, fn GroupByFunction) (EntityListMap, error)
 	Iterate(ctx context.Context, fn IteratorFunction) error
 }
-
```
dtfranz, line 30: More a question than a comment/critique: Why are these implementation methods no longer needed? Is the interface sufficiently implemented in other areas?

perdasilva: They are still needed. I've just  removed the GetContent from the EntitySource and made the EntityQuerier the EntiySource interface, basically. Just to simplify things. If a GetContent method is needed, one could always implement an EntitySource with it.

Issue #76, opened 2022-12-07
Create entity source implementation with registry v1 client


awgreene:
User Story:

As a Deppy User, I would like to use Deppy to consider if a set of constraints can be met by content within a catalogSource.

**Background:**

- The goal of the first milestone is to install/uninstall an operator.
- The operator that the milestone 1 installs will ultimately need to come from an existing catalogSource, as the alternatives would require implementing a new storage API for an operator, which was deemed too much work for milestone 1.
- The team has opted to wrap the existing registry client within a "Deppy Source Interface", constraining the Deppy library to only sourcing content from catalogSource but allowing us to reach the e2e milestone quickly.
- This approach will also allow us to write most of the code that will eventually be used by the "DeppySource catalogSource adapter" to convert existing content into the  format desired by Deppy.

**Acceptance Criteria:**

- Deppy can be provided with a catalogSource, treat it as an EntitySource, and retrieve the bundles and store them in a cache.
- Deppy can be notified that a catalogSource has been updated, retrieve the new content, and update the cache.
- Deppy can be notified that a catalogSource has been deleted and remove that source and it's entities from the cache.
- A test is implemented ensuring that Deppy can communicate with catalogSources.

**Out of Scope:**
- Using the registry client in the operator controller.

awgreene:
This is being moved to Milestone 2b, for the purposes of Milestone 1 we are returning a [hardcoded list](https://github.com/operator-framework/operator-controller/pull/101/files#diff-710ac95fcad0f796cb590b46a0f37d870f389ece5bab9d6aa0ee0165811335aaR8-R51)

PR #84, opened 2023-01-17
EntitySource implementation for the v1 registry client

ankitathomas:
An implementation of a RegistryEntitySource that watches CatalogSources on all namespaces. It maintains an internal cache 

See https://github.com/operator-framework/deppy/issues/76


pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+	}
+	if len(errs) > 0 {
+		return errors.NewAggregate(errs)
+	}
+	return nil
+}
+
+// imageIDFromCatalogSource returns the current image ref being run by a catalogSource
+func (r *CachedRegistryEntitySource) imageIDFromCatalogSource(ctx context.Context, catalogSource *v1alpha1.CatalogSource) string {
```
tmshort, line 205: Possibly return an `error` from this function, in addition to a `string`? Might make error handling easier, especially since you're collecting an array of errors.

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,257 @@
...
+				}()
+			}
+		}
+	}
+}
+
+func (r *CachedRegistryEntitySource) StopCache() {
+	r.Mutex.Lock()
+	defer r.Mutex.Unlock()
+	r.done <- struct{}{}
```
perdasilva, line 112: you should simply `close(r.done)`. That should be sufficient to trigger any `select case <-r.done: statement` ^^

pkg/deppy/input/catalogsource/registry_bundle_to_entity.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func entityFromBundle(catsrcID string, bundle *catalogsourceapi.Bundle) (*input.Entity, error) {
+	properties := map[string]string{}
+	var errs []error
+
+	propsList := map[string]map[string]struct{}{
+		property.TypeGVK:         {},
+		property.TypeGVKRequired: {},
+		property.TypePackage:     {},
```
perdasilva, line 30: Are we missing `TypePackageRequired`?

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,257 @@
...
+			if filter(entity) {
+				resultSet = append(resultSet, *entity)
+			}
+		}
+	}
+	return resultSet, nil
+}
+
+func (r *CachedRegistryEntitySource) GroupBy(ctx context.Context, fn input.GroupByFunction) (input.EntityListMap, error) {
+	r.Mutex.Lock()
```
perdasilva, line 143: I think you should use a RWLock and when reading the cache use RLock. You can have many RLocks, but only one (RW)Lock.

pkg/deppy/input/catalogsource/registry_bundle_to_entity.go
perdasilva: Oh, another thing that came to mind, we also need to know the default channel for the package of the bundle. This important for when we sort the bundles. When no channel is specified, we'd give priority to the default channel.

ankitathomas: Added both, we should now have access to both.

pkg/deppy/input/catalogsource/registry_bundle_to_entity.go
```diff
@@ -0,0 +1,189 @@
...
+)
+
+type UpgradeEdge struct {
+	property.Channel
+	Replaces  string   `json:"replaces,omitempty"`
+	Skips     []string `json:"skips,omitempty"`
+	SkipRange string   `json:"skipRange,omitempty"`
+}
+
+func entityFromBundle(catsrcID string, bundle *catalogsourceapi.Bundle) (*input.Entity, error) {
```
perdasilva: it might make sense to break this function down into sub-functions to make it a bit smaller and more readable. You may also want to make a type declaration for the key of the property map, e.g. `type NamespacedName string` or some such.

pkg/deppy/input/catalogsource/registry_grpc_client.go
```diff
@@ -0,0 +1,71 @@
...
+	"time"
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	catalogsourceapi "github.com/operator-framework/operator-registry/pkg/api"
+
+	"github.com/operator-framework/deppy/pkg/deppy/input"
+	"github.com/operator-framework/deppy/pkg/lib/grpc"
+)
+
+type RegistryClient interface {
```
perdasilva, line 17: can't we straight up use the registry client coming from operator-registry?

perdasilva: oh nvm - we do

pkg/deppy/input/catalogsource/registry_bundle_to_entity.go
```diff
@@ -0,0 +1,161 @@
...
+	setPropertyValue := func(key, value string) {
+		if _, ok := propsList[key]; !ok {
+			propsList[key] = map[string]struct{}{}
+		}
+		if _, ok := propsList[key][value]; !ok {
+			propsList[key][value] = struct{}{}
+		}
+	}
+
+	for _, prvAPI := range bundle.ProvidedApis {
```
perdasilva, line 51: It might be good to break these steps into their own functions to make this function a bit smaller and more readable. wdyt?

pkg/deppy/input/catalogsource/registry_bundle_to_entity.go
```diff
@@ -0,0 +1,161 @@
...
+
+	properties[TypeDefaultChannel] = pkg.DefaultChannelName
+	properties[TypeBundleSource] = bundle.BundlePath
+
+	if len(errs) > 0 {
+		return nil, fmt.Errorf("failed to parse properties for bundle %s/%s in %s: %v", bundle.GetPackageName(), bundle.GetVersion(), catsrcID, errors.NewAggregate(errs))
+	}
+
+	// Since multiple instances of bundle may exist for different channels, entityID must include reference to channel
+	entityIDFromBundle := func(catsrcID string, bundle *catalogsourceapi.Bundle) deppy.Identifier {
```
perdasilva, line 156: should we just push this out into a proper function rather than an inline one?

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+
+type Option func(*CachedRegistryEntitySource)
+
+func WithSyncInterval(d time.Duration) Option {
+	return func(c *CachedRegistryEntitySource) {
+		c.syncInterval = d
+	}
+}
+
+func NewCachedRegistryQuerier(client client.WithWatch, rClient RegistryClient, logger *logr.Logger, options ...Option) *CachedRegistryEntitySource {
```
perdasilva, line 47: could the registry client be internal to this struct?

perdasilva: might also be worth adding the logger as an option with `zap.New()` as the default value.
if you want to allow for custom registry client implementations, it might also be worth moving this to an option and using yours as the default.

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+		//imageID: imageID,
+	}
+	if !isManagedCatalogSource(catalogSource) {
+		r.queue.Forget(catalogSourceKey)
+		r.queue.AddAfter(catalogSourceKey, r.syncInterval)
+	}
+	r.logger.Info("Completed cache update", "catalogSource", catalogSourceKey)
+}
+
+func (r *CachedRegistryEntitySource) Get(ctx context.Context, id deppy.Identifier) *input.Entity {
```
perdasilva, line 163: rather than iterating over the whole cache to find the item, could you break out the catalog source component of the key and go directly to the right map?

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+		for _, entity := range entries.Items {
+			if entity.Identifier() == id {
+				return entity
+			}
+		}
+	}
+	return nil
+}
+
+func (r *CachedRegistryEntitySource) Filter(ctx context.Context, filter input.Predicate) (input.EntityList, error) {
```
perdasilva, line 176: you could re-use `Iterate` here to reduce /reuse some of the code

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+		for _, entity := range entries.Items {
+			if filter(entity) {
+				resultSet = append(resultSet, *entity)
+			}
+		}
+	}
+	return resultSet, nil
+}
+
+func (r *CachedRegistryEntitySource) GroupBy(ctx context.Context, fn input.GroupByFunction) (input.EntityListMap, error) {
```
perdasilva, line 190: same here - you could re-use `Iterate`

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+			catalogSource := v1alpha1.CatalogSource{}
+			err = json.Unmarshal(encodedObj, &catalogSource)
+			if err != nil {
+				r.logger.Error(err, "cannot reconcile non-catalogSource: unmarshalling failed", "event", entry.Type, "object", entry.Object)
+				continue
+			}
+
+			switch entry.Type {
+			case watch.Deleted:
+				func() {
```
perdasilva, line 101: I'd say move this to its own function, wdyt?

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+		r.logger.Info("Populated initial cache")
+	}
+	// watching catalogSource for changes works only with OLM managed catalogSources.
+	go r.ProcessQueue(ctx)
+	for {
+		select {
+		case <-r.done:
+			return
+		case entry := <-catalogSourceWatch.ResultChan():
+			encodedObj, err := json.Marshal(entry.Object)
```
perdasilva, line 87: another option might to be just cast it to a `*v1alpha1.CatalogSource`, e.g.
```
catalogsource, ok := entry.Object.(*v1alpha1.CatalogSource)
```

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+		return
+	}
+	if err := r.populate(ctx); err != nil {
+		r.logger.Error(err, "error populating initial entity cache")
+	} else {
+		r.logger.Info("Populated initial cache")
+	}
+	// watching catalogSource for changes works only with OLM managed catalogSources.
+	go r.ProcessQueue(ctx)
+	for {
```
perdasilva, line 82: might be worth wrapping this into its own function and using adding `defer r.Stop()` + any other shutdown bits

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+				r.logger.Error(err, "cannot reconcile non-catalogSource: unmarshalling failed", "event", entry.Type, "object", entry.Object)
+				continue
+			}
+
+			switch entry.Type {
+			case watch.Deleted:
+				func() {
+					r.RWMutex.Lock()
+					defer r.RWMutex.Unlock()
+					catalogSourceKey := types.NamespacedName{Namespace: catalogSource.Namespace, Name: catalogSource.Name}
```
perdasilva, line 104: maybe push this to its own function. I think I've seen it around a couple of times in the code

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+					delete(r.cache, catalogSourceKey.String())
+					r.logger.Info("Completed cache delete", "catalogSource", catalogSourceKey)
+				}()
+			case watch.Added, watch.Modified:
+				r.syncCatalogSource(ctx, catalogSource)
```
perdasilva, line 109: we should consider using the workqueue to drive sync in a different go routine. When something is added or modified add that resource to the workqueue. This could also help us with error handling and retries. E.g. if there's a connection failure or transient issue, we can just re-enqueue the resource after some time.

careful with modified because it can happen frequently - e.g. everytime the connection state changes - we should only trigger a resync for certain changes to the resource. Maybe when the connection state is READY?

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+	for {
+		item, _ := r.queue.Get() // block till there is a new item
+		defer r.queue.Done(item)
+		if _, ok := item.(types.NamespacedName); ok {
+			var catalogSource v1alpha1.CatalogSource
+			if err := r.client.Get(ctx, item.(types.NamespacedName), &catalogSource); err != nil {
+				r.logger.Info("cannot find catalogSource, skipping cache update", "CatalogSource", item)
+				return
+			}
+			r.syncCatalogSource(ctx, catalogSource)
```
perdasilva, line 134: we should catch any errors here and re-enqueue the catalog if it they happen. We may want to understand how this work queue works so we can have an exponential backoff. It may also be worth emitting a kube event that says that the cache couldn't be sync'ed for a particular catalog and that it will retry in so many seconds.

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		r.logger.Error(err, "failed to list entities for catalogSource entity cache update", "catalogSource", catalogSourceKey)
+		if !isManagedCatalogSource(catalogSource) {
+			r.queue.AddRateLimited(catalogSourceKey)
+		}
+		return
+	}
+	r.cache[catalogSourceKey.String()] = sourceCache{
+		Items: entities,
+		//imageID: imageID,
```
perdasilva, line 154: is this comment still valid?

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+	defer r.RWMutex.Unlock()
+	entities, err := r.rClient.ListEntities(ctx, &catalogSource)
+	if err != nil {
+		r.logger.Error(err, "failed to list entities for catalogSource entity cache update", "catalogSource", catalogSourceKey)
+		if !isManagedCatalogSource(catalogSource) {
+			r.queue.AddRateLimited(catalogSourceKey)
+		}
+		return
+	}
+	r.cache[catalogSourceKey.String()] = sourceCache{
```
perdasilva, line 152: nice - change the whole entry at once - makes it atomic ^^

pkg/deppy/input/catalogsource/registry_cache_querier.go
```diff
@@ -0,0 +1,248 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	"github.com/operator-framework/deppy/pkg/deppy"
+	"github.com/operator-framework/deppy/pkg/deppy/input"
+)
+
+const defaultCatalogSourceSyncInterval = 5 * time.Minute
+
+type CachedRegistryEntitySource struct {
+	sync.RWMutex
```
perdasilva, line 25: how come this is exported rather than an internal variable? is this a Go pattern?

pkg/deppy/input/catalogsource/registry_grpc_client.go
```diff
@@ -0,0 +1,78 @@
...
+	if grpcTimeout == 0 {
+		grpcTimeout = grpc.DefaultGRPCTimeout
+	}
+	return &registryGRPCClient{timeout: grpcTimeout}
+}
+
+func (r *registryGRPCClient) ListEntities(ctx context.Context, catalogSource *v1alpha1.CatalogSource) ([]*input.Entity, error) {
+	// TODO: create GRPC connections separately
+	conn, err := grpc.ConnectWithTimeout(ctx, catalogSource.Address(), r.timeout)
+	if conn != nil {
```
perdasilva, line 35: nit: you could change this to

```
if err != nil {
  return ...
}
defer conn.Close()
```

pkg/deppy/input/catalogsource/registry_grpc_client.go
```diff
@@ -0,0 +1,78 @@
...
+		pkg, ok := catalogPackages[packageKey]
+		if !ok {
+			pkg, err = catsrcClient.GetPackage(ctx, &catalogsourceapi.GetPackageRequest{Name: bundle.PackageName})
+			if err != nil {
+				return entities, fmt.Errorf("failed to get package %s: %v", bundle.PackageName, err)
+			}
+			catalogPackages[packageKey] = pkg
+		}
+
+		entity, err := EntityFromBundle(catalogSourceID, pkg, bundle)
```
perdasilva, line 71: since this function is in the same package we can get away with not exporting it - or did you export it to make testing easier?

pkg/deppy/input/catalogsource/registry_grpc_client.go
```diff
@@ -0,0 +1,78 @@
...
+		if err != nil {
+			return entities, fmt.Errorf("failed to parse entity %s: %v", entity.Identifier(), err)
```
perdasilva, line 73: do you mean to return partial results? it might be worth just returning nil

pkg/deppy/input/catalogsource/registry_grpc_client_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+import (
+	"context"
+	"net"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	api2 "github.com/operator-framework/operator-registry/pkg/api"
```
perdasilva, line 11: maybe registryapi rather than api2?

perdasilva: or registry

pkg/deppy/input/catalogsource/registry_grpc_client_test.go
```diff
@@ -0,0 +1,177 @@
...
+			ID: "//prometheus/beta/0.37.0",
+			Properties: map[string]string{
+				"olm.gvk":                    `[{"group":"monitoring.coreos.com","kind":"Prometheus","plural":"prometheuses","version":"v1"},{"group":"monitoring.coreos.com","kind":"Prometheus","version":"v1"},{"group":"monitoring.coreos.com","kind":"PrometheusRule","plural":"prometheusrules","version":"v1"},{"group":"monitoring.coreos.com","kind":"PrometheusRule","version":"v1"}]`,
+				"olm.package":                `{"packageName":"prometheus","version":"0.37.0"}`,
+				"olm.channel":                `{"channelName":"beta","priority":0}`,
+				"olm.package.defaultChannel": "beta",
+				"olm.bundle.path":            "quay.io/openshift-community-operators/prometheus@sha256:6fbd3eaa123054c5023323d1f9ab7cbea178087fcb7cb4f3e83872c6a88d39a1",
+			},
+		}}))
+	})
```
perdasilva, line 176: might be worth adding error test cases as well

pkg/lib/grpc/grpc.go
```diff
@@ -0,0 +1,125 @@
+package grpc
```
perdasilva, line 1: doesn't need to be in this PR, but I think we should open a ticket and move all of this to the registry repo. It seems like it could be useful beyond this use-case

pkg/lib/grpc/grpc.go
```diff
@@ -0,0 +1,125 @@
...
+				return nil, err
+			}
+			return dialer.Dial("tcp", addr)
+		}))
+	}
+
+	return grpc.Dial(address, dialOptions...)
+}
+
+func grpcProxyURL(addr string) (*url.URL, error) {
```
perdasilva, line 86: I feel like I've seen this code before. Is it from the olm repo? if we move this to operator-registry we should also update olm.

pkg/deppy/input/catalogsource/registry_cache_querier_test.go
```diff
@@ -0,0 +1,360 @@
...
+func (r *fakeRegistryClient) ListEntities(ctx context.Context, catsrc *v1alpha1.CatalogSource) ([]*input.Entity, error) {
+	catalogSourceKey := types.NamespacedName{Namespace: catsrc.Namespace, Name: catsrc.Name}.String()
+	if src, ok := r.catalogSource[catalogSourceKey]; ok {
+		return src.Entities, src.err
+	}
+	return []*input.Entity{}, nil
+}
+
+var _ = Describe("Registry EntitySource", func() {
+	var querier input.EntitySource
```
perdasilva, line 48: maybe just make it a straigh up RegistryCacheQuerier so you don't need to cast it back in the AfterEach?

pkg/deppy/input/catalogsource/registry_bundle_to_entity_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/operator-framework/deppy/pkg/deppy/input/catalogsource"
+
+	catalogsourceapi "github.com/operator-framework/operator-registry/pkg/api"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("RegistryBundleConverter", func() {
+	It("generates entity from bundle", func() {
```
perdasilva, line 14: you may want to also poke at some of the error cases

perdasilva (changes requested):
Looks really good and like you've put a lot of effort into this. Thank you!
I think we should move all of this code to the `operator-controller` and remove the dependencies on olm and registry from deppy.
I've left a few comments that I hope will help ^^ Please let me know if you need any help or have any questions =D

ankitathomas:
Closing in favor of https://github.com/operator-framework/operator-controller/pull/129

PR #85, opened 2023-01-18
refactor constraints

perdasilva:
This PR updates the constraint definition to export the constraint structs. This enables constraint introspection, type switching, and overloading, which are helpful when trying to decipher the resolver errors as well as to allow us to provide more meaningful error messages. E.g. instead of saying "x is mandatory", you could say "x is installed", or "you're trying to install x, but..."

This PR also fixes some of the unit test files that were missing the ginkgo hook to run


pkg/deppy/constraint/constraint.go
```diff
@@ -9,45 +9,63 @@ import (
...
-func (constraint mandatory) Anchor() bool {
+func (constraint *MandatoryConstraint) Anchor() bool {
 	return true
 }
 
 // Mandatory returns a Constraint that will permit only solutions that
 // contain a particular Variable.
 func Mandatory() deppy.Constraint {
-	return mandatory{}
+	return &MandatoryConstraint{}
```
tmshort, line 51: Why is this not returning a `*deppy.Constraint`?

pkg/deppy/constraint/constraint.go
```diff
@@ -9,45 +9,63 @@ import (
...
 	return lm.LitOf(subject)
 }
 
-func (constraint mandatory) Order() []deppy.Identifier {
+func (constraint *MandatoryConstraint) Order() []deppy.Identifier {
 	return nil
 }
 
-func (constraint mandatory) Anchor() bool {
+func (constraint *MandatoryConstraint) Anchor() bool {
```
tmshort, line 44: Do these methods (including the ones above and below) need to use `*MandatoryConstraint`? The methods don't modify the constraint. The original methods did not use pointers.

tmshort: This applies to a good chunk of the constraint functions.

perdasilva: I tend to use pointer receivers because they are more efficient in moving data around. And since we'll be moving variables (together with their constraints) around often, I figured this would be an improvement.

tmshort: "Learning Go" states that pointers are not necessarily faster, and that using pointers should only be for methods that modify the struct/interface. That being said, as a hard-core C programmer, I find it hard to believe that copying would potentially be faster...

perdasilva: If that's the case, then I don't have strong feelings one way or the other. I tend to think more C like in this cases and try to use pointers rather than copy stuff around. I would say pointer received isn't necessarily wrong. If you'd prefer a value received, I'm happy to refactor ^^

awgreene: There's also a memory impact when not using pointers.

awgreene: I think you'd need to introduce performance testing on a case by case basis if you want to identify the speed tradeoffs.

tmshort: > There's also a memory impact to consider when not using pointers.

The implication in this case (since it's effectively an argument to the function) is that it's put on the stack, so the heap is not involved. Optimally, they would be put into registers if the structure were small enough.

pkg/deppy/constraint/constraint_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+func TestPkg(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Constraint")
+}
+
+var _ = Describe("Constraint", func() {
+	Describe("UserFriendlyConstraint", func() {
+		It("should provide the custom constraint message", func() {
+			userFriendlyConstraint := constraint.NewUserFriendlyConstraint(constraint.Mandatory(), func(constraint deppy.Constraint, subject deppy.Identifier) string {
```
awgreene, line 23: lol

pkg/deppy/constraint/constraint_test.go
```diff
@@ -0,0 +1,29 @@
...
+				return fmt.Sprintf("'%s' just _has_ to be there or you can't even...", subject)
```
awgreene, line 24: lol

awgreene (approved):
Nice work on this, constraints are much clearer.

tmshort:
Looks like there are some failures.

pkg/deppy/constraint/constraint.go
tmshort: Still not fixed?

PR #86, opened 2023-01-23
Refactor solver solution struct

perdasilva:
This PR refactors the `Solution` struct to be a little more helpful. It now returns returns the resolution error (if any), the selected variables, given a variable ID you can see whether it has been selected, and with an option, all variables used in the resolution can also be returned.

### DIFF
--- a/pkg/deppy/solver/solver.go
+++ b/pkg/deppy/solver/solver.go
@@ -2,6 +2,7 @@ package solver
 
 import (
 	"context"
+	"errors"
 
 	"github.com/operator-framework/deppy/internal/solver"
 	"github.com/operator-framework/deppy/pkg/deppy"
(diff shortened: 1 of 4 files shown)


pkg/deppy/solver/solver.go
```diff
@@ -11,10 +12,56 @@ import (
...
-// Solution is returned by the Solver when a solution could be found
-// it can be queried by Identifier to see if the entity was selected (true) or not (false)
-// by the solver
-type Solution map[deppy.Identifier]bool
+// Solution is returned by the Solver when the internal solver executed successfully.
+// A successful execution of the solver can still end in an error when no solution can
+// be found.
+type Solution struct {
+	err       deppy.NotSatisfiable
+	selection map[deppy.Identifier]deppy.Variable
```
dtfranz, line 20: What is the difference between the `deppy.Variable`s in the `selection` map vs the `variables` list? Is `selection` a subset of `variables`?

pkg/deppy/solver/solver_test.go
```diff
@@ -89,14 +89,13 @@ var _ = Describe("Entity", func() {
...
-			deppy.Identifier("2"): Equal(true),
-			deppy.Identifier("3"): Equal(false),
+		Expect(solution.SelectedVariables()).To(MatchAllKeys(Keys{
+			deppy.Identifier("1"): Equal(input.NewSimpleVariable("1", constraint.Mandatory(), constraint.Dependency("2"))),
+			deppy.Identifier("2"): Equal(input.NewSimpleVariable("2")),
 		}))
 	})
 
-	It("should fail when a dependency is prohibited", func() {
+	It("should place resolution errors should in the solution", func() {
```
dtfranz, line 98: "should place resolution errors in the solution"?

pkg/deppy/solver/solver_test.go
```diff
@@ -105,8 +104,18 @@ var _ = Describe("Entity", func() {
 		s := NewEntitySource(variables)
 		so, err := solver.NewDeppySolver(s, s)
 		Expect(err).To(BeNil())
-		_, err = so.Solve(context.Background())
-		Expect(err).Should(HaveOccurred())
+		solution, err := so.Solve(context.Background())
+		Expect(err).Should(BeNil())
```
dtfranz, line 108: Nitpick: I like how `Expect(err).ToNot(HaveOccurred())` reads as opposed to `BeNil()`. It's also more consistent with some of the other statements, like on the next line.

pkg/deppy/solver/solver.go
perdasilva: That's right - the selection are the variables that were selected by the solver. `variables` are _all_ the variables that were considered by the solver.

perdasilva: I've added some comments to clarify ^^

pkg/deppy/solver/solver_test.go
perdasilva: fixed ^^

pkg/deppy/solver/solver_test.go
perdasilva: nice one! Fixed ^^

perdasilva: Also - fixed it everywhere in the file XD

pkg/deppy/solver/solver.go
dtfranz: Makes sense, thanks!

dtfranz (approved):

PR #88, opened 2023-01-25
[pipeline pt1] Pipeline events

perdasilva:
This PR kicks off the pipeline package with the event subpackage.
Events are emitted by pipeline nodes. There are two types of events:
* Data events (the data type is pinned to the pipeline)
* Error events

Events have a header that include:
* an unique event id
* who created the event
* when did they create it
* who did they send it to 
* was it a broadcast event?
* any additional metadata that the user want to communicate

### DIFF
--- a/pkg/deppy/pipeline/api.go
+++ b/pkg/deppy/pipeline/api.go
@@ -1,0 +1,49 @@
+package pipeline
+
+import (
+	"time"
+)
+
+type EventSourceID string
+
+type EventType string
+type EventID string
+type EventMetadata map[string]string
+
+type EventHeader interface {
+	EventID() EventID
+	Creator() EventSourceID
+	Sender() EventSourceID
+	Receiver() EventSourceID
+	IsBroadcastEvent() bool
+	CreationTime() time.Time
+	Metadata() EventMetadata
+}
+
+type EventIDProvider interface {
+	NextEventID() EventID
+}
+
+type Event interface {
+	Header() EventHeader
+	Broadcast()
+	Route(dest EventSourceID)
+	String() string
+}
+
+type DataEvent[D interface{}] interface {
+	Event
+	Data() D
+	Copy() DataEvent[D]
+}
+
+type ErrorEvent interface {
+	Event
+	Error() error
+	Copy() ErrorEvent
+}
+
+type EventFactory[I interface{}] interface {
+	NewErrorEvent(err error) ErrorEvent
+	NewDataEvent(data I) DataEvent[I]
+}
--- a/pkg/deppy/pipeline/event/event_test.go
+++ b/pkg/deppy/pipeline/event/event_test.go
@@ -1,0 +1,103 @@
+package event_test
+
+import (
+	"encoding/json"
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/operator-framework/deppy/pkg/deppy/pipeline"
+	"github.com/operator-framework/deppy/pkg/deppy/pipeline/event"
+)
+
+var _ pipeline.EventIDProvider = &fakeCustomEventIDProvider{}
+
+type fakeCustomEventIDProvider struct {
+	fn func() pipeline.EventID
+}
+
+func (f fakeCustomEventIDProvider) NextEventID() pipeline.EventID {
+	return f.fn()
+}
+
+func TestEventFactory_NewDataEvent(t *testing.T) {
+	factory := event.NewEventFactory[int]("node")
+	dataEvent := factory.NewDataEvent(1)
+	assert.Equal(t, 1, dataEvent.Data())
+	assert.Equal(t, pipeline.EventSourceID("node"), dataEvent.Header().Creator())
+	assert.Nil(t, dataEvent.Header().Metadata())
+	assert.Empty(t, dataEvent.Header().Receiver())
+	assert.False(t, dataEvent.Header().IsBroadcastEvent())
+	assert.Equal(t, pipeline.EventSourceID("node"), dataEvent.Header().Sender())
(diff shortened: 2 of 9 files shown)


pkg/deppy/pipeline/event/eventidprovider/uuid.go
```diff
@@ -0,0 +1,41 @@
...
+
+var _ pipeline.EventIDProvider = &UUIDEventIDProvider{}
+
+type UUIDProviderFn func() (uuid.UUID, error)
+
+type UUIDEventIDProvider struct {
+	nextUUIDFn UUIDProviderFn
+}
+
+func NewUUIDEventIDProvider() *UUIDEventIDProvider {
```
varshaprasad96, line 21: I'm kind off confused on `EventProvider`'s implementation. Could you please explain the use of EventIDProvider that gives us the `nextIDFn`?
I was assuming the next event ID would be set by the node on the event. For which it may have to refer the routing table (not sure if routing table is the right word), but basically a cache that stores the overall state of each node, and the capabilities of each node. That is nodeID and what step of resolution it performs.

pkg/deppy/pipeline/event/header.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/operator-framework/deppy/pkg/deppy/pipeline"
+)
+
+var _ pipeline.EventHeader = &eventHeader{}
+
+type eventHeader struct {
```
varshaprasad96, line 12: Do we also want a field for event status. Not sure if header is the right place though. Basically event is either in `processed`, `processing` or `error` state. 

I thought of having status for the node, but then having a status on event should also be fine? Because an errored node can be re-triggered, but there is no such possibility with an event.

pkg/deppy/pipeline/api.go
```diff
@@ -0,0 +1,49 @@
...
+	"time"
+)
+
+type EventSourceID string
+
+type EventType string
+type EventID string
+type EventMetadata map[string]string
+
+type EventHeader interface {
```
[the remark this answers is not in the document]
perdasilva: I originally had a visited list with nodes and timestamps. But, eventually I removed it because in the DAG the events would be routed to their final destination, rather than go through the whole DAG. This is because of a peculiarity of the resolver that it needs _all_ the variables. So, you'd go from the required package variable source directly to the resolver node + to nodes that take those variables as input. But don't flow beyond those destination nodes. dykwim?

pkg/deppy/pipeline/event/event_test.go
```diff
@@ -0,0 +1,103 @@
...
+	factory := event.NewEventFactory[int]("node")
+	evt := factory.NewDataEvent(1)
+	evt.Broadcast()
+	assert.True(t, evt.Header().IsBroadcastEvent())
+}
+
+func TestEventFactory_Route(t *testing.T) {
+	factory := event.NewEventFactory[int]("node")
+	evt := factory.NewDataEvent(1)
+	evt.Route("node2")
```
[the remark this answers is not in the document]
perdasilva: The way I imagined it is each node knows where they need to send their output events to, which could be to multiple nodes. When the underlying process produces and event, the node will clone it and route it to each of its output nodes. This means just pushing these events to the nodes output channel. Listening to this channel is the router, which knows where to find the input channel for a node.

pkg/deppy/pipeline/event/eventidprovider/uuid.go
perdasilva: I just wanted a centralized way to generate unique event IDs. This implementation just uses a uuid library underneath with some error handling that will still give you a reasonably random and unique id. It's possible that we should add some mutex locks here just to make sure that two concurrent calls with the same error don't generate clashing uuids.

Another option could be to have an atomic counter. Then we can have monotonically increasing event ids.

Does this help? Basically, when we use the factory to create a new event, it already comes out-of-the-box with an unique id.

pkg/deppy/pipeline/event/header.go
perdasilva: idk that it would be necessary. At least, not now. Especially because events aren't really stored anywhere.

perdasilva:
@varshaprasad96 are you good with my answers? do you think we're gtg?

varshaprasad96:
@perdasilva I don't mind if we want to merge this PR as is, since its the first step in defining the skeleton of "Events" and its extensions. As we move ahead with implementation, we could answer a few more questions and come back with necessary changes. It would be difficult to get all the pipeline APIs perfect in one go, so we can keep iterating and making necessary changes based on how implementation takes us.

pkg/deppy/pipeline/api.go
```diff
@@ -0,0 +1,49 @@
...
+	Event
+	Data() D
+	Copy() DataEvent[D]
+}
+
+type ErrorEvent interface {
+	Event
+	Error() error
+	Copy() ErrorEvent
+}
```
varshaprasad96, line 44: @perdasilva Could we have this as:

```go
type Event[I interface{}] interface {
	Header() EventHeader
	Broadcast()
	Route(dest EventSourceID)
	String() string
	Content() I
}

type DataEvent[D interface{}] interface {
	Event[D]
	Copy() DataEvent[D]
}

type ErrorEvent interface {
	Event[error]
	Copy() ErrorEvent
}
```

This way we can have a method to get the content of the Event no matter what type it is. If its error event, then we could get the content to be error type for others it would be any type as specified. This is because, while working on tests in #89 (specifically [here](https://github.com/operator-framework/deppy/pull/89/files#diff-063708694d29a185cc28e09693d944055757f4eea2abfc9f8ef37c5c99ad7664R163)) it would be nice to be able to access the data of the event directly than convert it to JSON or string.

PR #89, opened 2023-03-14
[Pipeline pt2] Introduce router and routes for events

varshaprasad96:
This PR adds the logic for routing events with the help of routing table. The routing table consists of eventSourceIDs and their specific routes. The helpers are useful in routing an event to the necessary destination.

This is based on the work done by @perdasilva in https://github.com/perdasilva/olmcli

### DIFF
--- a/pkg/deppy/pipeline/route/router_test.go
+++ b/pkg/deppy/pipeline/route/router_test.go
@@ -2,7 +2,6 @@ package route
 
 import (
 	"context"
-	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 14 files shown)


pkg/deppy/pipeline/route/router_test.go
```diff
@@ -0,0 +1,279 @@
...
+			Expect(connForNode2).NotTo(BeNil())
+
+			noData = true
+			select {
+			case receivedMsg, exists := <-connForNode2.inputChannel:
+				Expect(receivedMsg.String()).To(ContainSubstring("testEvent"))
+				Expect(exists).To(BeTrue())
+			default:
+				// Shouldn't be entering default.
+				noData = false
```
varshaprasad96, line 212: I'm doubtful if this is the best way to write ginkgo tests for go channels. The non-occurring case everywhere is just being tested with a boolean for now.

PR #90, opened 2023-04-12
[EntitySource] Modify entity source interface

varshaprasad96:
The "Get" function of the entity source did not return an error and instead only a nil entity when its not found. This PR modifies the same, to make it return an error.

Reason:
The method could be implmented with a client that fetches bundle contents from the cluster. The error returned from the client needs to be propagated. Even if its from an offline key-value store, it could be upto the library users to ignore the returned error. 

More info: https://github.com/operator-framework/operator-controller/pull/141#discussion_r1163271945

### DIFF
--- a/pkg/deppy/input/cache_entity_source.go
+++ b/pkg/deppy/input/cache_entity_source.go
@@ -2,6 +2,7 @@ package input
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/operator-framework/deppy/pkg/deppy"
 )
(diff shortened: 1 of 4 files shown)


joelanford:
Looks good to me, but I'll defer to @perdasilva to make sure we aren't missing some important context about this.

ankitathomas (approved):
/lgtm

everettraven (approved):
/lgtm

perdasilva (approved):
Lgtm

joelanford (approved):

joelanford:
(Something seems off with the merge requirement for someone from `operator-framework/deppy-maintainers`. Me, @perdasilva, @ankitathomas , and @everettraven have all approved, and apparently none of us are on that team. I'm going to use my admin powers to merge anyway.

PR #95, opened 2023-05-02
[revise] Remove Entity and EntitySource

varshaprasad96:
Removes Entity and EntitySource. The solver takes in the Variable Source directly. The constraints, filters, sort or any transformation should be performed on variables before providing it to solver.


pkg/ext/olm/constraints.go
```diff
@@ -2,20 +2,19 @@ package olm
...
-
 	"github.com/operator-framework/deppy/pkg/deppy/input"
 
-	"github.com/operator-framework/deppy/pkg/deppy/constraint"
-
 	"github.com/operator-framework/deppy/pkg/deppy"
 )
 
+// Question:
+// These are not needed since while constructing the variables, its upto the users
```
varshaprasad96, line 13: @perdasilva is my understanding right?

PR #99, opened 2023-06-05
WIP: Preference ordering optional

perdasilva:


internal/solver/solve.go
```diff
@@ -145,4 +163,8 @@ var defaults = []Option{
 		}
 		return nil
 	},
+	func(s *solver) error {
+		s.disableOrderPreference = true
```
stevekuznetsov, line 167: I know this is pre-existing to your code, but ... why do we do:

```go
func NewSolver(options ...Option) (Solver, error) {
	s := solver{g: gini.New()}
	for _, option := range append(options, defaults...) {
		if err := option(&s); err != nil {
			return nil, err
		}
	}
	return &s, nil
}
```

By appending the defaults to the end, doesn't that overwrite passed-in options? The rest of the defaults have to go out of their way to check to see if there was user input first before doing something, but you haven't, and you couldn't unless you used `*bool`. Seems pretty convoluted. Also, why do we need closures for all of these? Seems like we'd be well served by:

```go
func NewSolver(options ...Option) (Solver, error) {
	litMap, err := newLitMapping(nil) // are variables really an *option* ?
	if err != nil {
		return nil, err
	}
	s := solver{
		g: gini.New(),
		litMap: litMap,
		tracer: DefaultTracer{},
		disableOrderPreference: true,
	}
	for _, option := range options {
		if err := option(&s); err != nil {
			return nil, err
		}
	}
	return &s, nil
}
```

pkg/deppy/input/cache/cache.go
```diff
@@ -0,0 +1,56 @@
+package cache
+
+import "sync"
+
+type Key string
+
+type Cache[I interface{}] interface {
```
stevekuznetsov, line 7:
```suggestion
type Cache[I any] interface {
```

pkg/deppy/input/cache/cache.go
```diff
@@ -0,0 +1,56 @@
+package cache
+
+import "sync"
+
+type Key string
```
stevekuznetsov, line 5: Can't this just be `comparable`? https://go.dev/blog/comparable

stevekuznetsov (commented):
I'm a bit lost as to why we're defining and implementing generic caches here. If what we want to do is to access a cache of k8s objects fed by a watch, we already have that implemented. Custom keys into that cache are added with indexers.

PR #100, opened 2023-06-09
Removes OLM extension from deppy

m1kola:
This is a domain specific package which should not be in Deppy.


perdasilva (approved):
Thank you!!!

varshaprasad96 (approved):
/lgtm
This looks good to me. We are already using some olm specific constraints (with respect to CRDUniqueness and BundleUniqueness) in operator-controller. I assume some of this code will also move there in case we start using it?

m1kola:
Yep, so some of the code was refactored and moved to operator-controller. Once/if we need something else - we will be able to use this as a reference/inspiration.

PR #103, opened 2023-06-16
Adds bingo for tool management

m1kola:

PR #104, opened 2023-06-16
Golangci lint upgrade

m1kola:
~~Includes https://github.com/operator-framework/deppy/pull/103 which needs to be merged first~~


.golangci.yaml
```diff
@@ -6,7 +6,6 @@ linters:
   enable:
   - asciicheck
   - bodyclose
-  - depguard
```
m1kola, line 9: `depguard` linter removed. We do not have configuration for it and new version of `golangci-lint` comes with the configuration where only Go standard library is allowed.

ncdc (approved):

PR #105, opened 2023-06-16
Adds extra linting rules

m1kola:


internal/solver/solve.go
```diff
@@ -35,16 +35,20 @@ const (
...
 // Context times out or is cancelled, an error is returned.
-func (s *solver) Solve(ctx context.Context) (result []deppy.Variable, err error) {
-	defer func() {
-		// This likely indicates a bug, so discard whatever
-		// return values were produced.
-		if derr := s.litMap.Error(); derr != nil {
-			result = nil
-			err = derr
-		}
-	}()
```
m1kola, line 46: `defer` which overrides return values is still a bit too smart. Took me a moment to understand how it works.

internal/solver/solve.go
```diff
@@ -35,16 +35,20 @@ const (
...
+func (s *solver) Solve(ctx context.Context) ([]deppy.Variable, error) {
+	result, err := s.solve(ctx)
```
ncdc, line 39: Please let me know what you think of this:

```go
result, err := s.solve(ctx)
if err != nil {
  return nil, err
}

if err := s.litMap.Error(); err != nil {
  return nil, err
}

return result, nil
```

m1kola: @ncdc  I'm not sure whether it is actually important here or not, but as far as I understand the original code the idea is that we return an error from `s.litMap.Error()` even if there was an error from `s.solve(ctx)`. In your code - we will be returning an error from `s.solve(ctx)` and ignoring error from `s.litMap.Error()`.

We can do this (use `return` instead of overriding variables):

```go
func (s *solver) Solve(ctx context.Context) ([]deppy.Variable, error) {
	result, err := s.solve(ctx)

	// This likely indicates a bug, so discard whatever
	// return values were produced.
	if derr := s.litMap.Error(); derr != nil {
		return nil, derr
	}

	return result, err
}
```

(we can use `err` instead of `derr` as well, but it is harder to follow and easier to mess up in the future, IMO).

ncdc: Right... I'd ask for @perdasilva to weigh in. I don't particularly like this structuring and would probably just keep the original defer structure with a named return variable in that case.

m1kola: @ncdc I really would like to avoid having named return and defer. Especially defer: too much magic in one place makes it hard to follow. I had to spend quite a few minutes to understand the order of returns.

ncdc (approved):

PR #131, opened 2023-08-01
Adds go module caching for bingo modules

m1kola:
We introduced [bingo](https://github.com/bwplotka/bingo) in OLMv1 repos for managing tools and it created a bunch of modules in the `.bingo` directory in each repo. It works well, but on every job we download these dependencies from internet instead of using cache. This happens becuase setup-go github action only looks into project root for `go.sum` and builds cache based on it.

`cache-dependency-path` needs to be added to all OLMv1 repos and it should speed up our builds (we've seen decrease in build time of ~3 minutes in https://github.com/operator-framework/olm-docs).

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -15,5 +15,8 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version-file: "go.mod"
+          cache-dependency-path: |
+            go.sum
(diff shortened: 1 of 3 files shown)


m1kola:
I'll close it for now to not confuse. It is not working as expected. More details [here](https://github.com/operator-framework/operator-controller/issues/323#issuecomment-1667920491).

PR #140, opened 2023-08-14
fix: solution.Error() returns untyped nil when SAT

### LINKED ISSUE #139: Using `solver.Solution.Error()` is awkward and unintuitive
I expected this to work correctly:
```go
if err := solution.Error(); err != nil {
  log.Println(err)
}
```

In reality, this code always logs "constraints not satisfiable", even the the solution was SAT.

To handle this correctly, I had to use this code instead:
```go
unsat := deppy.NotSatisfiable{}
if errors.As(solution.Error(), &unsat); len(unsat) > 0 {
	log.Println(unsat.Error())
}
```

We should fix `solution.Error()` to return an untyped `nil` error if the solution was SAT.

joelanford:
Fixes #139

### DIFF
--- a/pkg/deppy/solver/solver.go
+++ b/pkg/deppy/solver/solver.go
@@ -16,7 +16,7 @@ import (
 // A successful execution of the solver can still end in an error when no solution can
 // be found.
 type Solution struct {
-	err       deppy.NotSatisfiable
+	err       error
 	selection map[deppy.Identifier]deppy.Variable
 	variables []deppy.Variable
 }
--- a/pkg/deppy/solver/solver_test.go
+++ b/pkg/deppy/solver/solver_test.go
@@ -69,6 +69,7 @@ var _ = Describe("Entity", func() {
 		so := solver.NewDeppySolver(s, s)
 		solution, err := so.Solve(context.Background())
 		Expect(err).ToNot(HaveOccurred())
+		Expect(solution.Error()).ToNot(HaveOccurred())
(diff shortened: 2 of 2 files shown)


pkg/deppy/solver/solver.go
```diff
@@ -24,6 +24,9 @@ type Solution struct {
 // Error returns the resolution error in case the problem is unsat
 // on successful resolution, it will return nil
 func (s *Solution) Error() error {
+	if s.err == nil {
+		return nil
+	}
```
m1kola, line 29: I think we should just use `error` interface for `s.err` instead a specific type. This way we avoid having this strange `if` and still can avoid nil not being nil issue.

```diff
diff --git a/pkg/deppy/solver/solver.go b/pkg/deppy/solver/solver.go
index d60036f..f8758f5 100644
--- a/pkg/deppy/solver/solver.go
+++ b/pkg/deppy/solver/solver.go
@@ -16,7 +16,7 @@ import (
 // A successful execution of the solver can still end in an error when no solution can
 // be found.
 type Solution struct {
-       err       deppy.NotSatisfiable
+       err       error
        selection map[deppy.Identifier]deppy.Variable
        variables []deppy.Variable
 }
```

joelanford: Done! Wasn't sure if that change would have knock-on effects. Looks like it compiles and tests pass though, so I assume all good.

m1kola (approved):

PR #160, opened 2023-11-15
⚠️ Remove variable sources and make Deppy accept a slice of variables instead

### LINKED ISSUE #153: Breaking change: remove variable sources and make Deppy accept a slice of variables instead
We recently [removed](https://github.com/operator-framework/deppy/pull/108) entity soruces API. I'm wondering whether we should get rid of concept of variable sources as well and make Deppy accept a slice of variables.

### LINKED ISSUE #154: Breaking change: Remove getter methods on constraints and expose fields instead
Today constraints hold important data in unexported fields and provide getters to access data.

It seems like there is not much value in these getters: Deppy itself doesn't seem to use them and on the user side they are normally not required in production code.

However this makes testing of components which generate variables with constraints harder for deppy users. Users have to use something like `go-cmp` with [`AllowUnexported`](https://pkg.go.dev/github.com/google/go-cmp@v0.6.0/cmp#AllowUnexported) & use constructors such as [`constraint.AtMost()`](https://github.com/operator-framework/deppy/blob/3a99ed74aa56e34e0a4beba81fe12e4b7dae4e06/pkg/deppy/constraint/constraint.go#L195-L203) to create constraints instead of simply just comparing literals.

I think we should explore exporting constraint fields and removing getter methods.

For example, `AtMostConstraint` which looks today like this:

https://github.com/operator-framework/deppy/blob/3a99ed74aa56e34e0a4beba81fe12e4b7dae4e06/pkg/deppy/constraint/constraint.go#L159-L177

will be turned into something like this:

```go
type AtMostConstraint struct { 
 	IDs []deppy.Identifier 
 	N   int 
 } 
  
 func (constraint *AtMostConstraint) String(subject deppy.Identifier) string { 
 	s := make([]string, len(constraint.IDs)) 
 	for i, each := range constraint.IDs { 
 		s[i] = string(each) 
 	} 
 	return fmt.Sprintf("%s permits at most %d of %s", subject, constraint.N, strings.Join(s, ", ")) 
 } 
```

m1kola:
Instead of a variable source make Deppy accept a slice of variables as an input.

Closes #153

### DIFF
--- a/pkg/deppy/solver/solver.go
+++ b/pkg/deppy/solver/solver.go
@@ -1,16 +1,14 @@
 package solver
 
 import (
-	"context"
 	"errors"
 
 	"github.com/operator-framework/deppy/internal/solver"
 	"github.com/operator-framework/deppy/pkg/deppy"
-	"github.com/operator-framework/deppy/pkg/deppy/input"
 )
 
 // TODO: should disambiguate between solver errors due to constraints
-//       and other generic errors (e.g. entity source not reachable, etc.)
+//       and other generic errors
 
 // Solution is returned by the Solver when the internal solver executed successfully.
 // A successful execution of the solver can still end in an error when no solution can
@@ -18,7 +16,6 @@ import (
 type Solution struct {
 	err       error
 	selection map[deppy.Identifier]deppy.Variable
-	variables []deppy.Variable
 }
 
 // Error returns the resolution error in case the problem is unsat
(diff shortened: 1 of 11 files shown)


pkg/deppy/solver/solver.go
```diff
@@ -1,24 +1,21 @@
...
+//       and other generic errors
 
 // Solution is returned by the Solver when the internal solver executed successfully.
 // A successful execution of the solver can still end in an error when no solution can
 // be found.
 type Solution struct {
 	err       error
 	selection map[deppy.Identifier]deppy.Variable
-	variables []deppy.Variable
 }
```
m1kola, line 19: I think we can get rid of `Solution` completely (not in this PR). I do not see a lot of value in it.

And probably replace `DeppySolver` wrapper around internal package by exporting what we currently have in internal pacakge (with some modifications and also not in this PR). 

This comment is for me to look into whether it makes sense and create issues for this.

pkg/deppy/solver/solver.go
```diff
@@ -116,9 +68,5 @@ func (d DeppySolver) Solve(ctx context.Context, options ...Option) (*Solution, e
 		solution.err = unsatError
 	}
 
-	if solutionOpts.addVariablesToSolution {
-		solution.variables = vars
-	}
```
m1kola, line 121: Previously variables were generated by a varaible soruce. So for cases where the calling code for some reason needed to access input variables - we had an option to add them into the `Solution` struct. Since now the calling code provides input variables - this seems like pointless option. So I removed it.

pkg/deppy/solver/solver_test.go
```diff
@@ -141,26 +116,6 @@ var _ = Describe("Entity", func() {
 		}))
 	})
 
-	It("should add all variables to solution if option is given", func() {
```
m1kola, line 144: No longer have this option.

pkg/deppy/solver/solver_test.go
```diff
@@ -107,32 +93,21 @@ var _ = Describe("Entity", func() {
...
-		so := solver.NewDeppySolver(varSrcStruct)
-
-		solution, err := so.Solve(context.Background())
+		so := solver.NewDeppySolver()
+		solution, err := so.Solve(variables)
 		Expect(err).ToNot(HaveOccurred())
 		Expect(solution.Error()).To(HaveOccurred())
 	})
 
-	It("should return peripheral errors", func() {
```
m1kola, line 118: No longer have a variable soruce to fail

m1kola:
I think this is ready, but I'll keep it in the draft because I want another PR to go in first so we can cut a release.

pkg/deppy/solver/solver.go
stevekuznetsov: Structuring the logic to be helpers in generating the inputs to and consuming the outputs of the solver library sounds like a really good idea - lets folks use as much of or as little of this helpful logic as they need and doesn't bind them to what we happen to have or not to have wrapped at some moment in time.

stevekuznetsov (approved):

everettraven (approved):
I think these changes look great and will make it easier to use, thanks @m1kola !

ncdc (approved):

Issue #161, opened 2023-11-16
Breaking change: remove `DeppySolver` wrapper and export internal solver


m1kola:
When we merge #160 (issue #154) which removes variable sources the `DeppySolver` becomes not very useful: just a thin wrapper around [the solver from the internal package](https://github.com/operator-framework/deppy/blob/ddaf504f3bafa717f26d4dc0cd4d16c1631acdfd/internal/solver/solve.go).

I think we can get rid of `DeppySolver` as well as [`Solution` struct](https://github.com/operator-framework/deppy/blob/ddaf504f3bafa717f26d4dc0cd4d16c1631acdfd/pkg/deppy/solver/solver.go#L15-L41).

I think [this](https://github.com/operator-framework/deppy/blob/ddaf504f3bafa717f26d4dc0cd4d16c1631acdfd/internal/solver/solve.go#L34-L48) should become something like below and get exported:

```go
// Solve takes a slice containing all Variables and returns a slice
// containing only those Variables that were selected for
// installation. If no solution is possible an error is returned.
func (s *solver) Solve(vars []deppy.Variable) ([]deppy.Variable, error) {
	// ...
}
```

PR #164, opened 2023-12-01
🌱 Update the internal solver before exporting

m1kola:
Related to #161

Preparing the internal solver to be exported and used without [the `DeppySolver` wrapper](https://github.com/operator-framework/deppy/blob/e676ccebc62ac3a514bf381d1647fc7d9da13bd0/pkg/deppy/solver/solver.go#L40-L72).

### DIFF
--- a/internal/solver/solve.go
+++ b/internal/solver/solve.go
@@ -1,7 +1,6 @@
 package solver
 
 import (
-	"context"
 	"errors"
 	"fmt"
 
@@ -12,17 +11,8 @@ import (
 	"github.com/operator-framework/deppy/pkg/deppy"
 )
 
-var ErrIncomplete = errors.New("cancelled before a solution could be found")
-
-type Solver interface {
-	Solve() ([]deppy.Variable, error)
-}
-
-type solver struct {
-	g      inter.S
-	litMap *litMapping
+type Solver struct {
 	tracer deppy.Tracer
-	buffer []z.Lit
 }
 
 const (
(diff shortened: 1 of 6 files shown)


everettraven (approved):
These changes look good to me. That being said, I'm not well versed with the entire system and interactions in deppy so another approval is probably good to have prior to merging (I don't think my approval actually counts for merging in this repo anyways)

internal/solver/solve.go
```diff
@@ -33,81 +23,87 @@ const (
...
 	var aset map[z.Lit]struct{}
 	// push a new test scope with the baseline assumptions, to prevent them from being cleared during search
-	outcome, _ := s.g.Test(nil)
+	outcome, _ := giniSolver.Test(nil)
 	if outcome != satisfiable && outcome != unsatisfiable {
 		// searcher for solutions in input Order, so that preferences
-		// can be taken into acount (i.e. prefer one catalog to another)
-		outcome, assumptions, aset = (&search{s: s.g, lits: s.litMap, tracer: s.tracer}).Do(context.Background(), assumptions)
+		// can be taken into account (i.e. prefer one catalog to another)
+		outcome, assumptions, aset = (&search{s: giniSolver, lits: litMap, tracer: s.tracer}).Do(assumptions)
```
ncdc, line 67: Instead of removing the context param from Do, would it make more sense to plumb a context param all the way up from here to Solver.Solve? Do we think solving needs to be stoppable?

m1kola: @ncdc I think in the current form where we accept a static set of inputs it doesn't make a lot of sense for solving to be stoppable.

I guess it was necessary when Deppy had an abstraction for variable and entity sources. There was an assumption that sources can be slow (e.g. fetching data from Kube API).

PR #165, opened 2023-12-01
Remove `DeppySolver` wrapper and export internal solver instead

### LINKED ISSUE #161: Breaking change: remove `DeppySolver` wrapper and export internal solver

m1kola:
Closes #161


m1kola:
@ncdc rebased. Please take another look.

ncdc (approved):

PR #166, opened 2023-12-04
Remove getter methods from constraints

### LINKED ISSUE #154: Breaking change: Remove getter methods on constraints and expose fields instead

m1kola:
We export fields from constraints instead

Closes #154

### DIFF
--- a/pkg/deppy/constraint/constraint.go
+++ b/pkg/deppy/constraint/constraint.go
@@ -82,15 +82,15 @@ func Not() deppy.Constraint {
 }
 
 type DependencyConstraint struct {
-	dependencyIDs []deppy.Identifier
+	DependencyIDs []deppy.Identifier
 }
 
 func (constraint *DependencyConstraint) String(subject deppy.Identifier) string {
-	if len(constraint.dependencyIDs) == 0 {
+	if len(constraint.DependencyIDs) == 0 {
 		return fmt.Sprintf("%s has a DependencyConstraint without any candidates to satisfy it", subject)
 	}
-	s := make([]string, len(constraint.dependencyIDs))
-	for i, each := range constraint.dependencyIDs {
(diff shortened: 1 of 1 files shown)


m1kola:
@ncdc rebased on top of `main` (not conflicts) and marked as ready. Please take another look.

ncdc (approved):

